PR #17, opened 2015-08-13
Adding 2015-08-01 Microsoft.Web API schema

naveedaz:


schemas/2015-08-01/Microsoft.Web.json
```diff
@@ -0,0 +1,389 @@
...
+                        "Microsoft.Web/sites/config",
+                        "config"
+                    ]
+                },
+                "apiVersion": {
+                    "enum": [
+                        "2015-08-01"
+                    ]
+                },
+                "name": {
```
wenwu449: Do you really need this? all three configs "web, connectionstring, appsettings" has name property already and they have different schema with this one.

schemas/2015-08-01/Microsoft.Web.json
```diff
@@ -0,0 +1,389 @@
...
+                            }
+                        }
+                    }
+                }
+            }
+        },
+        "appsettings": {
+            "type": "object",
+            "properties": {
+                "name": {
```
wenwu449: Is property "name" required? if so please set "required".

schemas/2015-08-01/Microsoft.Web.json
```diff
@@ -0,0 +1,389 @@
...
+        "web": {
+            "type": "object",
+            "properties": {
+                "name": {
+                    "pattern": "web"
+                },
+                "properties": {
+                    "type": "object",
+                    "properties": {
+                        "phpVersion": {
```
wenwu449: is any of the property required?

schemas/2015-08-01/Microsoft.Web.json
naveedaz: With the name property enum here, the intellisense is better. The experience is that the user sees 3 choices for name, instead of just an empty string. Together with the name patterns on the web, connectionstrings and appsettings.schemas, it pivots the intellisense onto the right schema. If user selects web, they only see web specific properties, same goes for the other 2.

schemas/2015-08-01/Microsoft.Web.json
naveedaz: None of them are required.

schemas/2015-08-01/Microsoft.Web.json
naveedaz: Technically not required, but probably good to have it in there so users have the right values. I'll change it

PR #19, opened 2015-08-14
SKU promoted to the ARM envelope, ServerFarm replaced with ServerFarmId

naveedaz:


schemas/2015-08-01/Microsoft.Web.json
```diff
@@ -17,48 +17,37 @@
                         "2015-08-01"
                     ]
                 },
-                "properties": {
+                "sku": {
                     "type": "object",
                     "properties": {
                         "name": {
```
wenwu449: name is required, right?

naveedaz: Yes. Thanks. I have fixed it now.

PR #21, opened 2015-08-17
Add schema for Microsoft.MySql.

wimoy:


yangl900:
you will also need add a link to the root deploymentTemplate.json file.

schemas/2015-09-01/Microsoft.MySql.json
```diff
@@ -0,0 +1,489 @@
...
+        "allowAzureServices": {
+          "type": "boolean"
+        },
+        "dailyBackupTimeInHour": {
+          "type": "integer"
+        },
+        "options": {
+          "type": "object",
+          "properties": {
+            "div_precision_increment": {
```
yangl900: we will discuss this, usually we don't use underscore in property names. But since we are mapping MySQL configurations here, probably it's ok.

PR #26, opened 2015-08-20
Correct json schema informational messages in VS2015

toddbrownmcis:
Items must be unique
schemas\2014-04-01-preview\deploymentParameters.json
Line 40

Reference could not be resolved
schemas\2015-08-01\Microsoft.Web.json
Lines 56, 189, 283, and 307


toddbrownmcis:
Cleaning up informational messages in Visual Studio 2015

wenwu449:
Thank you for the fix!

PR #30, opened 2015-09-01
update frontendIPConfigurations to array

sjkp:
I believe that frontendIPConfigurations property on the loadBalancers properties should have been an array like the other properties. At least that is what is working when you use the schema to deploy a loadbalancer.

### DIFF
--- a/schemas/2015-08-01/Microsoft.Network.json
+++ b/schemas/2015-08-01/Microsoft.Network.json
@@ -276,14 +276,14 @@
             ]
         },
         "frontendIPConfigurations": {
-            "properties": {
-                "name": {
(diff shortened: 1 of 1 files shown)


schemas/2015-08-01/Microsoft.Network.json
```diff
@@ -276,14 +276,14 @@
             ]
         },
         "frontendIPConfigurations": {
-            "properties": {
-                "name": {
-                    "type": "string"
-                },
                 "properties": {
```
wenwu449: please fix the tab space

wenwu449:
Thank you for the fix.

PR #36, opened 2015-09-23
Adding a resource meta-schema and fixing the description property location for child resources.

daschult:
This pull request is bringing in the concept of a resource meta-schema that existing resource schemas can be validated against. This should help to make the resource schemas more consistent and it should encourage new resource providers to write schemas.
The resource meta-schema enforces the following rules:
- All resource schemas must have "id", "$schema", "title", "description", and "resourceDefinitions" properties.
- All resources and first level resource properties require non-empty descriptions.
- All resources must have an "apiVersion" property with a set of enumerated string values that match the yyyy-mm-dd(-preview) format.


daschult:
I'm going to review this more inside my own team first, then I'll submit this work for review to this repository.

PR #39, opened 2015-09-30
Add schema for Microsoft.Cache REST API 2015-08-01.

TimLovellSmith:


schemas/2015-08-01/Microsoft.Cache.json
```diff
@@ -0,0 +1,157 @@
+﻿{
+    "id": "http://schema.management.azure.com/schemas/2015-08-01/Microsoft.Cache.json",
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "title": "Microsoft.Cache",
+    "description": "Microsoft Cache Resource Types",
+    "definitions": {
+        "virtualNetworkId": {
+            "type": "string",
+            "pattern": "^/subscriptions/[^/]*/resourceGroups/[^/]*/providers/Microsoft.ClassicNetwork/virtualNetworks/[^/]*$",
+            "description": "The resource id of an Azure Classic Virtual Network"
```
philon-msft: "ID"

schemas/2015-08-01/Microsoft.Cache.json
```diff
@@ -0,0 +1,157 @@
...
+                                                "Basic",
+                                                "Standard",
+                                                "Premium"
+                                            ]
+                                        },
+                                        {
+                                            "$ref": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression"
+                                        }
+                                    ],
+                                    "description": "Microsoft.Cache/Redis: sku/name"
```
philon-msft: Convention for descriptions like this is to include the path to the parent object, then a description of this property. This one should be something like "Microsoft.Cache/Redis/sku: Name of the Redis SKU"

schemas/2015-08-01/Microsoft.Cache.json
```diff
@@ -0,0 +1,157 @@
...
+                    ]
+                },
+                "properties": {
+                    "type": "object",
+                    "properties": {
+                        "sku": {
+                            "type": "object",
+                            "properties": {
+                                "name": {
+                                    "oneOf": [
```
philon-msft: Can you provide defaults for this or any of the other properties?

TimLovellSmith: This is a schema, not a deployment template right? Just because you put a default value in the schema doesn't mean the deployment someone does is actually going to have that default value applied, does it?

philon-msft: Correct - the default just lets JSON editors suggest a default for people who are authoring a template using your schema. The default value won't end up in the deployment unless the user chooses that default in their schema. 
For an example, see how these other RP schemas specify defaults: Microsoft.AppService.json, Microsoft.Scheduler.json, Microsoft.Cache.json.

TimLovellSmith: @philon-msft Thanks for explaining, I hadn't realized putting defaults in here might actually have some effect.

schemas/2015-08-01/Microsoft.Cache.json
TimLovellSmith: @philon-msft So is 'properties' meant to be omitted here? Such that it's not necessary to say Microsoft.Cache/Redis/properties/sku? Then the parent path for something under 'properties' will just be Microsoft.Cache/Redis is that right?

philon-msft: @TimLovellSmith Correct, "properties" is just part of the JSON schema syntax. Users editing templates won't see it, so it shouldn't appear in descriptions.

schemas/2015-08-01/Microsoft.Cache.json
TimLovellSmith: added

schemas/2015-08-01/Microsoft.Cache.json
```diff
@@ -0,0 +1,160 @@
...
+                                            "$ref": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression"
+                                        }
+                                    ],
+                                    "description": "Microsoft.Cache/Redis/sku: along with capacity, determines capacity of the cache",
+                                    "default": "C"
+                                },
+                                "capacity": {
+                                    "oneOf": [
+                                        {
+                                            "type": "integer"
```
daschult: The JSON Schema Draft 4 specification defines a type called "positiveInteger" that might be useful here instead of "integer".

schemas/2015-08-01/Microsoft.Cache.json
```diff
@@ -0,0 +1,160 @@
...
+                            "required": [
+                                "name",
+                                "family",
+                                "capacity"
+                            ]
+                        },
+                        "shardCount": {
+                            "oneOf": [
+                                {
+                                    "type": "integer"
```
daschult: The JSON Schema Draft 4 specification defines a type called "positiveInteger" that might be useful here instead of "integer".

schemas/2015-08-01/Microsoft.Cache.json
```diff
@@ -0,0 +1,160 @@
...
+                                {
+                                    "$ref": "#/definitions/virtualNetworkId"
+                                },
+                                {
+                                    "$ref": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression"
+                                }
+                            ],
+                            "description": "Microsoft.Cache/Redis (optional) the ARM resource ID of a Classic Virtual Network for the redis cache to join"
+                        },
+                        "subnet": {
```
daschult: I believe subnets are a child-resource of virtualNetworks, right? In that case, could you get rid of "virtualNetwork" and just ask the deployment template author to provide a subnet resource uri, which would contain the virtual network?

TimLovellSmith: While that sounds like a good design there were some practical reasons we did the API like this, and the api-version is done now. I think for the sake of this PR we should only consider whether this schema is reasonable given the current API.

schemas/2015-08-01/Microsoft.Cache.json
TimLovellSmith: I don't think it's worth changing it. I would expect tools will actually be better at supporting integer than positiveInteger, and chance of user error here seems minimal.

schemas/2015-08-01/Microsoft.Cache.json
```diff
@@ -0,0 +1,160 @@
+﻿{
+    "id": "http://schema.management.azure.com/schemas/2015-08-01/Microsoft.Cache.json",
```
wenwu449: To expose this to template author, please add the reference to 2014-04-01-preview\deploymentTemplate.json and 2015-01-01\deploymentTemplate.json

schemas/2015-08-01/Microsoft.Cache.json
TimLovellSmith: ditto: I don't think it's worth changing it. I would expect tools will actually be better at supporting integer than positiveInteger, and chance of user error here seems minimal.

TimLovellSmith:
I've referenced the new files from deploymentTemplate.json and merged latest master. Please review again.

TimLovellSmith:
@wenwu449 Thank you!

PR #40, opened 2015-10-01
Fix various minor issues

philon-msft:
These are fixes for issues I discovered while evaluating the template authoring experience in editors that support schema validation.

### DIFF
--- a/schemas/2015-08-01/Microsoft.Compute.json
+++ b/schemas/2015-08-01/Microsoft.Compute.json
@@ -67,7 +67,8 @@
                     "type": "string"
                 },
                 "sku": {
-                    "type": "string"
+                    "type": "string",
+                    "default": "latest"
                 },
                 "version": {
                     "type": "string"
--- a/schemas/2015-08-01/Microsoft.Compute.json
+++ b/schemas/2015-08-01/Microsoft.Compute.json
@@ -67,11 +67,11 @@
                     "type": "string"
                 },
                 "sku": {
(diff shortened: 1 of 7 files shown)


daschult:
Looks good to me.

OptikSage:
[X] LGTM

schemas/2015-08-01/Microsoft.Compute.json
```diff
@@ -62,7 +67,8 @@
                     "type": "string"
                 },
                 "sku": {
-                    "type": "string"
+                    "type": "string",
+                    "default": "latest"
```
wenwu449: "latest" is the default value of "version", right?

philon-msft: Yes, that's the intent of this change

schemas/2015-08-01/Microsoft.Web.json
```diff
@@ -374,7 +386,11 @@
                         }
                     }
                 }
-            }
+            },
+            "required": [
+              "name",
```
wenwu449: can you also add "type" and "apiVersion" as required?

philon-msft: There are no "type" or "apiVersion" properties on this object. Unless I misunderstand your comment?

wenwu449: my bad, must be lost in different objects/types. You are right.

PR #44, opened 2015-10-17
Schema changes to support new networking features

FrankSiegemund:
Changes to template, compute and network schemas to fix bugs and to support multiple networking features:
1. Route support
2. Multi-NIC support in Compute schema
3. Support new DNS features
4. Support scaleset LB features
5. Fixing issues


FrankSiegemund:
@DeepakRajendranMsft @clamanna @IlyaGrebnov @mahthi

schemas/2015-08-01/Microsoft.Compute.json
```diff
@@ -42,6 +42,26 @@
                 "id"
             ]
         },
+        "networkInterfaces": {
+            "properties": {
+                "id": {
+                    "type": "string"
+                },
+                "properties": {
```
wenwu449: "primary" should be under another "properties"

schemas/2015-08-01/Microsoft.Network.json
```diff
@@ -35,7 +46,11 @@
                         "2015-06-15"
                     ]
                 },
+                "name": {
```
wenwu449: This is not necessary, all ARM resources require "name" property.

wenwu449: "name" property is not necessary, and it should be under "properties" if we need it. 
All ARM resources require "type", "name", "apiVerision" properties, we have this requirement defined in deploymentTemplate.json.

schemas/2015-08-01/Microsoft.Network.json
```diff
@@ -44,31 +59,34 @@
...
-                    "type": "object",
                     "required": [
                         "publicIPAllocationMethod"
                     ]
                 }
             },
             "required": [
+                "type",
+                "apiVersion",
+                "name",
```
wenwu449: Not necessary.

schemas/2015-08-01/Microsoft.Network.json
```diff
@@ -294,42 +334,27 @@
...
+                "properties"
             ]
         },
         "backendAddressPools": {
+            "type": "object",
             "properties": {
                 "name": {
                     "type": "string"
-                },
-                "properties": {
```
wenwu449: So the "name" will be the only property of this object? VS will have a hard time to match a object to this definition because almost all objects here has a "name" property, if "name" is the only property, you can add ""additionalProperties": false," to help VS on matching.

schemas/2015-01-01/deploymentTemplate.json
```diff
@@ -60,6 +60,7 @@
                                     { "$ref": "http://schema.management.azure.com/schemas/2015-08-01/Microsoft.Network.json#/definitions/loadBalancers" },
                                     { "$ref": "http://schema.management.azure.com/schemas/2015-08-01/Microsoft.Network.json#/definitions/networkSecurityGroups" },
                                     { "$ref": "http://schema.management.azure.com/schemas/2015-08-01/Microsoft.Network.json#/definitions/networkInterfaces" },
+                                    { "$ref": "http://schema.management.azure.com/schemas/2015-08-01/Microsoft.Network.json#/definitions/routeTables" },
```
wenwu449: Please update 2014-04-01-preview\deploymentTemplate.json as well.

schemas/2015-08-01/Microsoft.Compute.json
FrankSiegemund: For some reason, this worked in VS. Sorry for the overlook. Fixed.

schemas/2015-01-01/deploymentTemplate.json
FrankSiegemund: Fixed.

schemas/2015-08-01/Microsoft.Network.json
FrankSiegemund: Yes, it is the only field.

schemas/2015-01-01/deploymentTemplate.json
```diff
@@ -416,4 +417,5 @@
             "description": "API version of the resource type"
         }
     }
-    }
+}
+
```
wenwu449: Please remove the extra line at the end of the file.

FrankSiegemund: Fixed.

PR #49, opened 2015-10-25
Add schema for KeyVault.

rhencke:


schemas/2015-06-01/Microsoft.KeyVault.json
```diff
@@ -0,0 +1,209 @@
+{
+    "id": "http://schema.management.azure.com/schemas/2015-06-01/Microsoft.KeyVault.json#",
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "title": "Microsoft.KeyVault",
+    "description": "Microsoft KeyVault Resource Types",
+    "definitions": {
+        "keyPermission": {
+            "anyOf": [
```
wenwu449: can we have "all" and "backup" in the same time for "keyPermission"? if not, please use "oneOf"

schemas/2015-06-01/Microsoft.KeyVault.json
```diff
@@ -0,0 +1,209 @@
...
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/definitions/secretPermission"
+                    },
+                    "uniqueItems": true,
+                    "minItems": 1,
+                    "description": "Microsoft.KeyVault/vaults: The permissions granted on secrets in this vault to this Active Directory object."
+                }
+            },
+            "minProperties": 1
```
wenwu449: consider to use "additionalProperties: false" to make sure at least one of the "keys" and "secrets" is there.

schemas/2015-06-01/Microsoft.KeyVault.json
```diff
@@ -0,0 +1,209 @@
...
+                            "type": "string",
+                            "pattern": "^[0-9a-fA-F]{8}(-[0-9a-fA-F]{4}){3}-[0-9a-fA-F]{12}$",
+                            "description": "Microsoft.KeyVault/vaults: The tenant ID of the Azure Active Directory Tenant to use for authorization."
+                        },
+                        "accessPolicies": {
+                            "type": "array",
+                            "items": {
+                                "$ref": "#/definitions/accessPolicy"
+                            },
+                            "enabledForDeployment": {
```
wenwu449: is this another property of "vaults"?

schemas/2015-01-01/deploymentTemplate.json
```diff
@@ -37,6 +37,7 @@
                             },
                             {
                                 "oneOf": [
+                                    { "$ref": "http://schema.management.azure.com/schemas/2015-06-01/Microsoft.KeyVault.json#/definitions/secrets" },
```
wenwu449: can you please make these change in 2014-04-01-preview/deploymentTemplate.json too?

schemas/2015-06-01/Microsoft.KeyVault.json
```diff
@@ -0,0 +1,209 @@
...
+                        "sku",
+                        "accessPolicies"
+                    ]
+                },
+                "resources": {
+                    "type": "array",
+                    "items": {
+                        "allOf": [
+                            { "$ref": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/resourceBase" },
+                            { "$ref": "#/definitions/secrets" }
```
wenwu449: Please create a definition called "secretChild" with type as "secrets", and reference here. 
"2014-04-01-preview/Microsoft.Sql.json#/definitions/databases" and "2014-04-01-preview/Microsoft.Sql.json#/definitions/databasesChild" are good examples.

schemas/2015-06-01/Microsoft.KeyVault.json
rhencke: Thank you for the review!

This is an area where I could use some guidance, honestly.  So, unfortunately, yes - you can.  But this seems to conflict with recommendations.  For example, the following is accepted (and stored verbatim) by the API:

```json
"keys": [
  "list",
  "all",
  "all",
  "all",
  "list"
],
"secrets": [
  "all",
  "list",
  "list"
]
```

While that is accepted by the KeyVault provider, the PowerShell cmdlets explicitly forbid mixing of 'all' and other levels of permissions:
```
PS C:\Users\Robert> Set-AzureKeyVaultAccessPolicy <# ... #> -PermissionsToKeys all, list
Set-AzureKeyVaultAccessPolicy : Invalid permissions to keys. The list contains "all" along
with other permissions. Please remove "all" from the list or include only "all" in the list.

PS C:\Users\Robert> Set-AzureKeyVaultAccessPolicy <# ... #> -PermissionsToSecrets all, list
Set-AzureKeyVaultAccessPolicy : Invalid permissions to keys. The list contains "all" along
with other permissions. Please remove "all" from the list or include only "all" in the list.
```

So I can either go the way PowerShell did, and try to prevent setting 'all' along with other permissions, or I can allow mixing 'all' with other access levels.  My initial thought was to try to provide a consistent development experience, so I tried to emulate what PowerShell would accept, but this does not represent in truth everything the provider allows.  Do you have a preference here?

schemas/2015-06-01/Microsoft.KeyVault.json
rhencke: D'oh.  Yes.  I messed this up when I was rearranging code.  Good catch - thank you.

schemas/2015-01-01/deploymentTemplate.json
rhencke: Will put this in.

schemas/2015-06-01/Microsoft.KeyVault.json
rhencke: I'll get this changed.  Thanks.

schemas/2015-06-01/Microsoft.KeyVault.json
rhencke: Thanks - I wasn't familiar with additionalProperties.  It looks like I should be using this a lot more (for example, it looks like using this would have caught the error you found below with enabledForDeployment).

schemas/2015-06-01/Microsoft.KeyVault.json
wenwu449: I would suggest to define schema according what provider accepts as valid inputs.
If you really want to go with the PowerShell way, you need to define "all" and other enum values as two objects and use "oneOf" to refernce them in Ln57.

schemas/2015-06-01/Microsoft.KeyVault.json
rhencke: Looking at this further, the provider currently accepts but ignores additional properties on every object but the root-level resource.  I'm going to see if I can find another way to enforce this.

schemas/2015-06-01/Microsoft.KeyVault.json
rhencke: Thanks.  I'll change this to be a simple array that accepts exactly what the provider accepts as valid inputs, then.

rhencke:
Hi @wenwu449.  Please take another look.

wenwu449:
@rhencke, we have some change recently and a "resourceDefinition" object is introduced, all the definitions will be exposed to template author are moved under that object.
Can you merge with the origin/master branch and update accordingly?

rhencke:
Hi @wenwu449 .  Please take another look.

The changes to 0e50f37 look much more chaotic than they really are.  resourceDefinitions was placed at the top to keep with convention in the other files, and the resources were alphabetized.  All changes should be purely structural here - the only content change should be at https://github.com/rhencke/azure-resource-manager-schemas/commit/0e50f37dd29711764417c163a60c8ca0baa2533c#diff-16753de7d94b4984677e601250191dbeL149 to change reference to ```#/resourceDefinitions/secretsChild```.

PR #50, opened 2015-10-26
Creating resourceDefinitions section in resource schemas

daschult:
This pull request introduces the concept of "resourceDefinitions" into our resource schema. Currently our resource schema only contain a "definitions" section where the definitions of our Azure resources were meant to go. However, as resource schema have been authored, many schema authors have found that several of their resources had shared sub-definitions, so they broke these sub-definitions out into their own definition JSON object that could be referenced by their resources. This created a situation where the "definitions" section in a resource schema now had both resource definitions and reusable generic definitions.
By introducing a "resourceDefinitions" section and moving all of our resource definitions from "definitions" to "resourceDefinitions", we can improve the readability of our resource schema. This will also make validating our resource schema (both automated and manual) much easier because it physically separates the resource definitions from the reusable definitions. Validators can now apply a different set of validation techniques to resource definitions than they apply to generic definitions.


schemas/2015-01-01/deploymentTemplate.json
```diff
@@ -37,35 +37,34 @@
...
-                                    { "$ref": "http://schema.management.azure.com/schemas/2014-06-01/Microsoft.Web.json#/definitions/certificates" },
-                                    { "$ref": "http://schema.management.azure.com/schemas/2014-06-01/Microsoft.Web.json#/definitions/serverfarms" },
-                                    { "$ref": "http://schema.management.azure.com/schemas/2015-08-01/Microsoft.Web.json#/definitions/certificates" },
-                                    { "$ref": "http://schema.management.azure.com/schemas/2015-08-01/Microsoft.Web.json#/definitions/serverfarms" },
-                                    { "$ref": "http://schema.management.azure.com/schemas/2014-04-01-preview/Microsoft.Sql.json#/definitions/databases" },
-                                    { "$ref": "http://schema.management.azure.com/schemas/2014-04-01-preview/Microsoft.Sql.json#/definitions/firewallrules" },
-                                    { "$ref": "http://schema.management.azure.com/schemas/2014-04-01/Microsoft.Insights.json#/definitions/alertrules" },
```
wenwu449: this line is removed, is it on purpose?

daschult: Nope. That wasn't on purpose at all. I'll put it back in.

daschult: Done.

schemas/2014-10-01-preview/microsoft.authorization.json
```diff
@@ -3,7 +3,7 @@
     "$schema": "http://json-schema.org/draft-04/schema#",
     "title": "Microsoft.Authorization",
     "description": "Microsoft Microsoft.Authorization Resource Types",
-    "definitions": {
+    "resourceDefinitions": {
```
wenwu449: "httpActionRequest" in this file is not a resource, we should keep it as "definitions"

schemas/2015-03-01-preview/Microsoft.AppService.json
```diff
@@ -172,31 +160,6 @@
...
-          ],
-          "description": "CSM resource name"
-        },
-        "resourceType": {
-          "type": "string",
-          "description": "Type of a CSM resource"
-        }
-      }
-    },
     "appServiceApiVersion": {
```
wenwu449: "apiServiceApiVersion" is not a resource, so it should be under "definitions", right?

daschult: Correct. Done.

schemas/2015-03-01-preview/Microsoft.AppService.json
```diff
@@ -14,7 +14,7 @@
           ]
         },
         "apiVersion": {
-          "$ref": "#/definitions/appServiceApiVersion"
+          "$ref": "#/resourceDefinitions/appServiceApiVersion"
```
wenwu449: reference need to be updated too.

daschult: Done.

PR #55, opened 2015-11-03
Schema for Traffic Manager profile

vladca:
For review only, please do not merge yet.


schemas/2015-11-01/Microsoft.Network.json
```diff
@@ -0,0 +1,159 @@
...
+                                },
+                                "ttl": {
+                                    "type": "integer",
+                                    "minimum": 30,
+                                    "maximum": 2147483647
+                                },
+                                "fqdn": {
+                                    "type": "string"
+                                },
+                                "required": [
```
wenwu449: outside of "properties"

schemas/2015-11-01/Microsoft.Network.json
```diff
@@ -0,0 +1,159 @@
...
+                                    "type": {
+                                         "enum": [
+                                              "Microsoft.Network/trafficManagerProfiles/azureEndpoints",
+                                              "Microsoft.Network/trafficManagerProfiles/externalEndpoints",
+                                              "Microsoft.Network/trafficManagerProfiles/nestedEndpoints"
+                                        ]
+                                    },
+                                    "properties": {
+                                        "type": "object",
+                                        "properties": {
```
wenwu449: any of the properties is required?

schemas/2015-11-01/Microsoft.Network.json
```diff
@@ -0,0 +1,159 @@
...
+                                        }
+                                    },
+                                    "required": [
+                                        "name",
+                                        "type",
+                                        "properties"
+                                    ]
+                                }
+                            },
+                            "additionalItems": true
```
wenwu449: this means any object can be element of the array, is this expected by your service?

schemas/2015-11-01/Microsoft.Network.json
```diff
@@ -0,0 +1,159 @@
...
+                    "additionalProperties": true
+                }
+            },
+            "required": [
+                "apiVersion",
+                "name",
+                "type",
+                "properties"
+            ],
+            "additionalProperties": true
```
wenwu449: do you have some extra input property not listed?

schemas/2015-11-01/Microsoft.Network.json
```diff
@@ -0,0 +1,159 @@
+{
+    "id": "http://schema.management.azure.com/schemas/2015-11-01/Microsoft.Network.json#",
```
wenwu449: a general comment is that all properties in the schema type are defined explicitly, this will cause any template expression be denied by VS, for example, "[parameters("foo")]" will be marked as error for a property value for a trafficmanager resource in template.
I suggest for some properties we expect user will use parameters or other expressions, use "oneOf" and "$ref": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression", you can take schemas/2015-08-01/Microsoft.Cache.json as example.

vladca: done for all integer and enum properties except apiVersion and type

schemas/2015-11-01/Microsoft.Network.json
vladca: moved outside of properties

schemas/2015-11-01/Microsoft.Network.json
vladca: some properties are required but which of them are depends on "trafficRoutingMethod" and "endpoints/type"; none of these properties are required in any and all cases

schemas/2015-11-01/Microsoft.Network.json
vladca: other objects are no expected by our service; do we want to err on the side of strictness? can a user by-pass schema validation if it is for some reason broken or out of date?

schemas/2015-11-01/Microsoft.Network.json
vladca: We have read-only properties that the user might carry-over when copy-pasting from a GET request (which is, I imagine how most templates get written) and which the service will ignore. Is your recommendation to be strict and not allow these?

schemas/2015-11-01/Microsoft.Network.json
wenwu449: if other objects are not expected, we should use "additionalItems": false. 
schema validation is just a client check, nothing blocks user send ARM a request with a broken template. ARM still has service side check though.

schemas/2015-11-01/Microsoft.Network.json
wenwu449: If template author set value on read-only properties in template, ARM service will send them in request body to RP, RP will ignore them. Wouldn't user be confused about why the value is not set successfully? Client side schema validation will mark the read-only properties as error so user can double check to understand it. Again, nothing blocks them to send the template containing resources with read-only properties to ARM, schema only do client side checking in VS during authoring template.

schemas/2014-04-01-preview/deploymentTemplate.json
```diff
@@ -122,6 +123,7 @@
                         },
                         {
                             "enum": [
+                                "global",
```
FrankSiegemund: Suggest not to have this at the top. Have it at the bottom. This should not be the default first value that people see because it is not valid for the vast majority of our resources.

schemas/2015-11-01/Microsoft.Network.json
```diff
@@ -0,0 +1,222 @@
...
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "title": "Microsoft.Network",
+    "description": "Microsoft Network Resource Types",
+    "resourceDefinitions": {
+        "trafficManagerProfiles": {
```
FrankSiegemund: Why is the indentation here 4 instead of 2 as in the other Microsoft.Network.json file?

schemas/2014-04-01-preview/deploymentTemplate.json
```diff
@@ -73,6 +73,7 @@
                                     { "$ref": "http://schema.management.azure.com/schemas/2015-08-01/Microsoft.Network.json#/resourceDefinitions/networkSecurityGroups" },
                                     { "$ref": "http://schema.management.azure.com/schemas/2015-08-01/Microsoft.Network.json#/resourceDefinitions/networkInterfaces" },
                                     { "$ref": "http://schema.management.azure.com/schemas/2015-08-01/Microsoft.Network.json#/resourceDefinitions/routeTables" },
+                                    { "$ref": "http://schema.management.azure.com/schemas/2015-11-01/Microsoft.Network.json#/resourceDefinitions/trafficManagerProfiles" },
```
FrankSiegemund: Why 11-01? Isn't 12-01 closer?

schemas/2015-11-01/Microsoft.Network.json
wenwu449: We unified all JSON files to 2 space indentation after this PR was requested.
Vlad, please help to reformat to 2 space indentation for this file. Thanks!

schemas/2014-04-01-preview/deploymentTemplate.json
wenwu449: note: file path should align with apiVersion, for easier discovery.

vladca:
I botched the merge from Azure:master, I'll fix it

schemas/2015-11-01/Microsoft.Network.json
vladca: Will do

vladca:
I'll submit a new PR instead of this broken one so that I don't pollute Azure:master history

PR #60, opened 2015-11-17
add supported apiVersion for storage

bmoore-msft:


schemas/2015-08-01/Microsoft.Storage.json
```diff
@@ -14,7 +14,8 @@
                 },
                 "apiVersion": {
                     "enum": [
-                        "2015-05-01-preview"
+                        "2015-05-01-preview",
+                        "2015-06-15"
```
philon-msft: Looks good to me.

PR #78, opened 2015-12-07
Schema for Traffic Manager Profile resource

vladca:
Also reviewed in https://github.com/Azure/azure-resource-manager-schemas/pull/55.


schemas/2015-11-01/Microsoft.Network.json
```diff
@@ -0,0 +1,222 @@
+{
+  "id": "http://schema.management.azure.com/schemas/2015-11-01/Microsoft.Network.json#",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Microsoft.Network",
+  "description": "Microsoft Network Resource Types",
+  "resourceDefinitions": {
+    "trafficManagerProfiles": {
```
wenwu449: Please add "description" for this object.

schemas/2015-11-01/Microsoft.Network.json
```diff
@@ -0,0 +1,222 @@
...
+      "type": "object",
+      "properties": {
+        "apiVersion": {
+          "enum": [
+            "2015-11-01"
+          ]
+        },
+        "name": {
```
wenwu449: Actually you do not need this, deploymentTemplate.json has defined that and make that "required".

wenwu449:
Please add a test in "\tests\2015-11-01\Microsoft.Network.tests.json"

schemas/2015-11-01/Microsoft.Network.json
```diff
@@ -0,0 +1,231 @@
...
+        "type": {
+          "enum": [
+            "Microsoft.Network/trafficManagerProfiles"
+          ]
+        },
+        "properties": {
+          "type": "object",
+          "properties": {
+            "profileStatus": {
+              "anyOf": [
```
vladca: Wenjun, I had to change all of the "oneOf" clauses to "anyOf" because the positive tests (see the test file) were returning this error: "Data is valid against more than one schema from \"oneOf\": indices 0 and 1".

schemas/2015-11-01/Microsoft.Network.json
```diff
@@ -0,0 +1,231 @@
...
+                {
+                  "enum": [
+                    "Enabled",
+                    "Disabled"
+                  ]
+                },
+                {
+                  "$ref": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression"
```
vladca: Allowing an expression in an "anyOf" clause basically bypasses the constraint on the enum (or integer elsewhere) - anything that doesn't is not "Enabled" or "Disabled" will fall under the "expression" and validate fine. Is there any way to specify that the result of the expression must respect the constraint of the enum (or integer elsewhere)?

tests/2015-11-01/Microsoft.Network.tests.json
```diff
@@ -0,0 +1,262 @@
...
+          "dataPath": "/properties/endpoints/0/type"
+        }
+      ],
+      "definition": "http://schema.management.azure.com/schemas/2015-11-01/Microsoft.Network.json#/resourceDefinitions/trafficManagerProfiles",
+      "json": {
+        "name": "TestProfile-1",
+        "type": "gibberish",
+        "apiVersion": "gibberish",
+        "properties": {
+          "profileStatus": "gibberish",
```
vladca: This doesn't cause a validation error!

schemas/2015-11-01/Microsoft.Network.json
wenwu449: Vlad, please still use "oneOf", we need exact one match here, instead of "at least one". The error may be an issue of our test tool, I will work with @daschult on it.

schemas/2015-11-01/Microsoft.Network.json
wenwu449: Please use "oneOf".

tests/2015-11-01/Microsoft.Network.tests.json
wenwu449: should report error after using "oneOf".

schemas/2015-11-01/Microsoft.Network.json
vladca: Agreed, switching back to one of, cheers

tests/2015-11-01/Microsoft.Network.tests.json
vladca: It doesn't, though, for the same reason that it didn't for anyOf: "gibberish" doesn't satisfy the enum branch but is considered an expression so no failure

wenwu449:
@vladca, Dan has fixed the issue in test tools, we also require all properties support "expression" now. With the latest test tools, there are a few errors on this schema, most because some properties not support "expression" yet, can you fix them?

vladca:
So really all properties, no exception? I pulled from master but I still don’t see Dan’s fix. I’d like to see the errors so I can check that I’ve really fixed them. Cheers

wenwu449:
Please use "git pull https://github.com/Azure/azure-resource-manager-schemas.git master" to merge latest test tools to your local branch.

schemas/2015-11-01/Microsoft.Network.json
```diff
@@ -0,0 +1,222 @@
...
+          "required": [
+            "trafficRoutingMethod",
+            "dnsConfig",
+            "monitorConfig"
+          ],
+          "additionalProperties": false
+        }
+      },
+      "required": [
+        "apiVersion",
```
wenwu449: Please include "location" here as well, if it is required.

PR #80, opened 2015-12-07
Adding expression support and more descriptions to all resource schemas

daschult:
Recently we found that many deployment templates were getting schema validation warnings because developers were using expressions ("[add(1, 2)]") instead of the type that the property required ("integer"). To solve this problem, we decided to go through the resource schemas and to add support for expressions to all properties.
I added validation to the ResourceMetaSchema.json file so that every property would be required to support expressions. I've fixed all of the errors that resulted from running "validateSchemas.js", but I'm sure that I missed some properties. This pull request was already getting large, so I decided to take it in chunks. This first pass is adding expression support for many of the properties and adding descriptions where they were missing. In the next pass I'm hoping to catch more of the properties that don't support expressions.

### DIFF
--- a/schemas/2014-02-26/microsoft.visualstudio.json
+++ b/schemas/2014-02-26/microsoft.visualstudio.json
@@ -21,11 +21,27 @@
           "type": "object",
           "properties": {
             "OperationType": {
-              "type": "string",
+              "anyOf": [
+                {
+                  "$ref": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json#/definitions/expression"
+                },
+                {
+                  "type": "string",
+                  "minLength": 1
+                }
+              ],
               "description": "microsoft.visualstudio/account: Type of operation being performed on the account, which can be either Create or Link."
             },
             "AccountName": {
-              "type": "string",
+              "anyOf": [
+                {
+                  "$ref": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json#/definitions/expression"
+                },
+                {
+                  "type": "string",
+                  "minLength": 1
+                }
+              ],
               "description": "microsoft.visualstudio/account: Name of the Visual Studio Online account"
             }
           }
@@ -70,11 +86,27 @@
           "type": "object",
           "properties": {
             "ProcessTemplateId": {
-              "type": "string",
+              "anyOf": [
+                {
+                  "$ref": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json#/definitions/expression"
+                },
+                {
+                  "type": "string",
+                  "minLength": 1
+                }
+              ],
               "description": "microsoft.visualstudio/account/project: Process template guid for the project"
             },
             "VersionControlOption": {
-              "type": "string",
+              "anyOf": [
+                {
+                  "$ref": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json#/definitions/expression"
+                },
+                {
+                  "type": "string",
+                  "minLength": 1
+                }
+              ],
               "description": "microsoft.visualstudio/account/project: Version control type for the project, currently TfsVc and Git"
             }
           }
--- a/schemas/2014-02-26/microsoft.visualstudio.json
+++ b/schemas/2014-02-26/microsoft.visualstudio.json
@@ -21,27 +21,13 @@
           "type": "object",
           "properties": {
             "OperationType": {
-              "oneOf": [
-                {
-                  "$ref": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json#/definitions/expression"
-                },
-                {
-                  "type": "string",
-                  "minLength": 1
-                }
-              ],
+              "type": "string",
+              "minLength": 1,
               "description": "microsoft.visualstudio/account: Type of operation being performed on the account, which can be either Create or Link."
             },
             "AccountName": {
-              "oneOf": [
-                {
-                  "$ref": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json#/definitions/expression"
-                },
-                {
-                  "type": "string",
-                  "minLength": 1
-                }
-              ],
+              "type": "string",
+              "minLength": 1,
               "description": "microsoft.visualstudio/account: Name of the Visual Studio Online account"
             }
           }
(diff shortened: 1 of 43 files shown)


schemas/2014-02-26/microsoft.visualstudio.json
```diff
@@ -21,11 +21,27 @@
           "type": "object",
           "properties": {
             "OperationType": {
-              "type": "string",
+              "anyOf": [
```
wenwu449: We should use "oneOf", and maybe not for "string" type, as any "string" will match.

schemas/2014-04-01-preview/Microsoft.Sql.json
```diff
@@ -364,11 +376,27 @@
       "type": "object",
       "properties": {
         "endIpAddress": {
-          "type": "string",
+          "oneOf": [
```
wenwu449: this may be an error because expression is a string.

schemas/2014-04-01-preview/Microsoft.Sql.json
```diff
@@ -364,11 +376,27 @@
...
+            {
+              "type": "string",
+              "minLength": 1
+            }
+          ],
           "description": "Microsoft.Sql/server/firewallrules: ending IP address"
         },
         "startIpAddress": {
-          "type": "string",
+          "oneOf": [
```
wenwu449: string and expression.

daschult: Fixed.

schemas/2014-04-01-preview/Microsoft.Sql.json
daschult: Fixed.

schemas/2014-04-01/Microsoft.Insights.json
```diff
@@ -244,53 +273,80 @@
...
+                },
+                {
+                  "type": "boolean"
+                }
+              ],
               "description": "Microsoft.Insights/webtests: Is the webtest enabled."
             },
             "Frequency": {
-              "type": "number",
+              "$ref": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json#/definitions/numberOrExpression",
```
wenwu449: not defined in 2014-04-01-preview one.

schemas/2014-04-01/Microsoft.Insights.json
```diff
@@ -244,53 +273,80 @@
...
-              "type": "array",
-              "items": {
-                "type": "object",
-                "properties": {
-                  "Id": {
-                    "type": "string",
-                    "description": "Microsoft.Insights/webtests: Location id of the webtest"
+              "oneOf": [
+                {
+                  "$ref": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json#/definitions/expression"
```
wenwu449: same issue.

tools/ResourceMetaSchema.json
```diff
@@ -1,108 +1,200 @@
...
-                        "properties": { }
-                    },
-                    "required": [ "apiVersion", "type" ],
-                    "additionalProperties": { "$ref": "#/definitions/resourceProperty" }
-                }
-            },
-            "required": [ "description", "properties" ]
+        "definitions": {
+          "type": "object",
+          "minProperties": 1,
```
wenwu449: this may not be true, I see some simple schema just has everything inline in resourceDefintions.

tools/ResourceMetaSchema.json
```diff
@@ -1,108 +1,200 @@
...
+      "required": [ "$schema", "title", "description", "resourceDefinitions" ],
+      "additionalProperties": false
+    },
+    "resource": {
+      "type": "object",
+      "properties": {
+        "type": { "enum": [ "object" ] },
+        "description": { "$ref": "#/definitions/description" },
+        "required": { "$ref": "#/definitions/stringArrayEmptyOkay" },
+        "properties": {
```
wenwu449: "properties" is required, I missed it last time.

schemas/2014-08-01/Microsoft.Scheduler.json
```diff
@@ -178,38 +243,58 @@
       "type": "object",
       "properties": {
         "uri": {
-          "type": "string",
-          "maxLength": 2048,
-          "description": "Microsoft.Scheduler/jobCollections/jobs: URI of the endpoint to call"
-        },
-        "method": {
           "anyOf": [
```
wenwu449: do we need this? If they take any string less than 2048 char?

tools/ResourceMetaSchema.json
```diff
@@ -1,108 +1,200 @@
...
-                        "type": { },
-                        "apiVersion": { "$ref": "#/definitions/apiVersionProperty" },
-                        "resources": { "$ref": "#/definitions/childResources" },
-                        "properties": { }
-                    },
-                    "required": [ "apiVersion", "type" ],
-                    "additionalProperties": { "$ref": "#/definitions/resourceProperty" }
-                }
-            },
-            "required": [ "description", "properties" ]
```
wenwu449: why remove the "properties" out of "required"?

tools/runSchemaTests.js
```diff
@@ -15,6 +15,7 @@ utilities.forEachFile(testsFolderPath, function (filePath) {
         testFiles.push(filePath);
```
wenwu449: need to update wiki according to this change.

tools/ResourceMetaSchema.json
```diff
@@ -1,108 +1,200 @@
...
+        "$ref": {
+          "type": "string",
+          "format": "uri"
         },
-        "description": {
-            "type": "string",
-            "minLength": 1
+        "items": { "$ref": "http://json-schema.org/draft-04/schema#/properties/items" },
+        "oneOf": { "$ref": "#/definitions/resourcePropertyDefinitionOptions" },
+        "type": { "enum": [ "string" ] },
```
wenwu449: this seems not correct. property under "properties" of a resource should be a object.

schemas/2014-04-01/Microsoft.Insights.json
daschult: Good catch. Fixed.

schemas/2014-04-01/Microsoft.Insights.json
daschult: Fixed.

tools/ResourceMetaSchema.json
daschult: This is not saying that "definitions" is required. It is saying that if "definitions" is declared, then it must not be empty.

schemas/2014-08-01/Microsoft.Scheduler.json
daschult: My thought was that if the expression ended up being longer than 2048 characters, then they would get schema errors. It's unlikely, but I decided it was possible.
If you feel the odds are too low to justify this, then I'll remove the expression clause of this "anyOf".

tools/ResourceMetaSchema.json
daschult: ...\azure-resource-manager-schemas\schemas\2015-06-01\Microsoft.KeyVault.json has resources that don't have "properties" as a resource-level property. That made me remove "properties" from the required list.

tools/ResourceMetaSchema.json
daschult: This was an observation of looking at the existing schemas. When we declare the type of a resource specific property, the "type" value is always "string".

tools/runSchemaTests.js
daschult: Done.

schemas/2014-08-01/Microsoft.Scheduler.json
wenwu449: OK, let's keep it.

PR #86, opened 2015-12-15
updated schema & tests

ryancrawcour:
new schema that enhances consistencyPolicy property
fixed tests that were failing
added new test for new consistencyPolicy

### DIFF
--- a/schemas/2015-04-08/Microsoft.DocumentDB.json
+++ b/schemas/2015-04-08/Microsoft.DocumentDB.json
@@ -46,30 +46,10 @@
             "consistencyPolicy": {
               "oneOf": [
                 {
-                  "type": "object",
-                  "properties": {
-                    "defaultConsistencyLevel": {
-                      "enum": [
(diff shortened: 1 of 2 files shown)


schemas/2015-04-08/Microsoft.DocumentDB.json
```diff
@@ -89,5 +69,67 @@
...
+              ]
+            },
+            { "$ref": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression" }
+          ]
+        }
+      },
+      "required": [
+        "defaultConsistencyLevel"
+      ]
+    }
```
daschult: I would recommend adding "additionalProperties: false" here after the "required" property so that if a user switches from using "BoundedStaleness" over to one of the other consistencyLevels, they are reminded that "maxStalenessPrefix" and "maxIntervalInSeconds" don't apply to these other levels.

PR #98, opened 2016-01-20
Updating "disablePasswordAuthentication" property type to boolean.

waynekuo:
linuxConfiguration/disablePasswordAuthentication should be a boolean property instead of string.


schemas/2015-08-01/Microsoft.Compute.json
```diff
@@ -594,7 +594,7 @@
     "linuxConfiguration": {
       "properties": {
         "disablePasswordAuthentication": {
-          "type": "string"
+          "type": "boolean"
```
wenwu449: can you add/update test for this change? Thanks

PR #129, opened 2016-03-29
Update network api version

DeepakRajendranMsft:
@johnmic  @FrankSiegemund


schemas/2015-08-01/Microsoft.Network.json
```diff
@@ -15,7 +15,8 @@
         "apiVersion": {
```
FrankSiegemund: You could add a test .json file under tests/

PR #133, opened 2016-03-31
Scheduler: fix API-version from 2014-08-01 to the correct 2014-08-01-…

pinwang81:
Fix the Scheduler API version issue.  Change it from 2014-08-01 to the correct API version 2014-08-01-preview.


schemas/2014-08-01-preview/Microsoft.Scheduler.json
```diff
@@ -1,5 +1,5 @@
 {
-  "id": "http://schema.management.azure.com/schemas/2014-08-01/Microsoft.Scheduler.json",
+  "id": "http://schema.management.azure.com/schemas/2014-08-01-preview/Microsoft.Scheduler.json",
```
wenwu449: Please update references in 2015-01-01\deploymentTemplate.json and 2014-04-01-preview\deploymentTemplate.json

PR #137, opened 2016-04-12
Dev/arroyc/arm schema

arroyc:
1. added platformUpdateDomainCount and platformFaultDomainCount validations for availabilitysets snippet
2. executed runSchemaTests.js successfully.

### DIFF
--- a/schemas/2015-08-01/Microsoft.Compute.json
+++ b/schemas/2015-08-01/Microsoft.Compute.json
@@ -22,11 +22,25 @@
           "type": "object",
           "properties": {
             "platformUpdateDomainCount": {
-              "$ref": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/numberOrExpression",
+              "oneOf": [
+                {
+                  "type": "number",
+                  "minimum": 1,
+                  "maximum": 20
+                },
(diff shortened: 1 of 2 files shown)


schemas/2015-08-01/Microsoft.Compute.json
```diff
@@ -22,11 +22,25 @@
           "type": "object",
```
wenwu449: Can you support expression for this "properties" too? (L21), currently it is failing ResourceMetaSchema.json validation.

tests/2015-08-01/Microsoft.Compute.tests.json
```diff
@@ -13,6 +13,96 @@
...
+          "platformFaultDomainCount": 0
+        }
+      }
+    },
+
+    {
+      "name": "availabilitySets - FaultDomainCount out of range - more",
+      "expectedErrors": [
+        {
+          "message": "Data does not match any schemas from \"oneOf",
```
wenwu449: There are tests failing now for runSchemaTests.js, mostly because the error message does not match. Please fix.

Error: The error message for error 1 was not the expected error.
Expected: "Data does not match any schemas from \"oneOf"
Actual:   "Data does not match any schemas from \"oneOf\""

TianoMS:
@arroyc I'll close this PR since it's too long ago. Please create a new PR if you still need these changes.

PR #146, opened 2016-04-28
Update Microsoft.Web.json

toddbrownmcis:
Fix setParameters property to be an object and allow an expression.  This was already fixed in 2014-06-01 version of Microsoft.Web.json.


naveedaz:
Looks good.

PR #153, opened 2016-05-19
Initial commit of Power BI deployment schema

JohnKurlak:
Initial commit of Power BI deployment schema.

### DIFF
--- a/schemas/2016-01-29/Microsoft.PowerBI.json
+++ b/schemas/2016-01-29/Microsoft.PowerBI.json
@@ -1,0 +1,71 @@
+{
+  "id": "http://schema.management.azure.com/schemas/2016-01-29/Microsoft.PowerBI.json",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Microsoft.PowerBI",
+  "description": "Microsoft Power BI Resource Types",
+  "resourceDefinitions": {
+    "workspaceCollections": {
+      "description": "Microsoft.PowerBI/workspaceCollections resource",
+      "type": "object",
+      "properties": {
+        "apiVersion": {
+          "enum": [
+            "2016-01-29"
+          ]
+        },
+        "type": {
+          "enum": [
+            "Microsoft.PowerBI/workspaceCollections"
+          ]
+        },
+        "name": {
+          "oneOf": [
+            {
+              "type": "string",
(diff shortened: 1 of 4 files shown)


daschult:
Do you have a swagger specification for this resource provider? If so, we can automatically generate this schema for you.

JohnKurlak:
Yes, we have a swagger specification at: https://github.com/Azure/azure-rest-api-specs/blob/5e432a8ba1a652997b304d6bbfcbf883aecc815f/arm-powerbiembedded/2016-01-29/swagger/powerbiembedded.json.

Can you please update the documentation on http://sharepoint/sites/AzureUX/Sparta/SpartaWiki/ARM%20Template%20Schema%20Publishing.aspx to indicate the possibility of using a Swagger doc (and what the steps for doing it are)?  It would save people a lot of time :)

schemas/2016-01-29/Microsoft.PowerBI.json
```diff
@@ -0,0 +1,101 @@
...
+              "pattern": "^[a-zA-Z0-9\\s]+$",
+              "minLength": 1,
+              "maxLength": 40
+            },
+            {
+              "$ref": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression"
+            }
+          ]
+        },
+        "tags": {
```
wenwu449: Please do not have the Azure common properties defined here as it is already defined in the root schema, reason:
1. if the pattern is updated in the root schema, it will be overwritten by RP schema
2. VS may evaluate the property improperly if the property was defined differently across RP schema and root schema.

schemas/2016-01-29/Microsoft.PowerBI.json
```diff
@@ -0,0 +1,101 @@
...
+              "$ref": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression"
+            }
+          ]
+        }
+      },
+      "required": [
+        "apiVersion",
+        "type",
+        "name",
+        "location",
```
wenwu449: please keep "location" in the required properties array if it is required for the resource type, even removing the property in the schema, because it has been already defined in root schema, but it is not required in the root schema because not all RP resource types require it.

JohnKurlak:
Updated (removed tags/location).

JohnKurlak:
Thanks so much!!!

PR #158, opened 2016-05-25
Add extra storage schema api versions generated from swagger documents

daschult:


wenwu449:
azure-resource-manager-schemas/schemas/2015-08-01/Microsoft.Storage.json already define 2015-05-01-preview and 2015-06-15 apiVersions, define them again may cause conflicts.
Can we only add the 2016-01-01 apiVersion one?

daschult:
Sure, I'll remove the older ones and only leave the latest.

daschult:
Removed.

PR #159, opened 2016-05-25
Add compute schemas generated from compute swagger documents

daschult:


wenwu449:
Similar to Microsoft.Storage schema, can we only update for the new apiVersion 2016-03-30?

daschult:
Removed.

schemas/2016-03-30/Microsoft.Compute.json
```diff
@@ -0,0 +1,1976 @@
...
+            }
+          ],
+          "description": "Gets or sets a list containing reference to all Virtual Machines created under this Availability Set."
+        },
+        "statuses": {
+          "oneOf": [
+            {
+              "type": "array",
+              "items": {
+                "$ref": "#/definitions/InstanceViewStatus"
```
yangl900: I thought this is readonly property?

daschult: It wasn't marked as "readOnly" in the Microsoft.Compute swagger document, so I have no way of knowing. I can remove it for now.

daschult:
Can we merge this pull request?

schemas/2016-03-30/Microsoft.Compute.json
```diff
@@ -0,0 +1,1962 @@
...
+            {
+              "$ref": "#/definitions/VirtualMachineProperties"
+            },
+            {
+              "$ref": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression"
+            }
+          ]
+        }
+      },
+      "resources": {
```
wenwu449: is this in the correct level? it should be under the "properties", right?

philon-msft:
@wenwu449, can this be merged now?

PR #161, opened 2016-05-25
Add generated Network resource schemas

daschult:


DeepakRajendranMsft:
@daschult, I have created a PR with tests. Some of the tests fail, I have added comments in there, could you please take a look.

https://github.com/daschult/azure-resource-manager-schemas/pull/2

tools/ResourceMetaSchema.json
```diff
@@ -195,8 +195,7 @@
             }
           },
           "required": [
-            "oneOf",
```
FrankSiegemund: Why is this part of this checkin? Is ARM OK with this? Having a description is probably what they want?

FrankSiegemund:
Comments:
1. The network schema addresses the same version as an existing file: zure-resource-manager-schemas\schemas\2015-08-01\Microsoft.Network.json. Both have 2016-03-30.

FrankSiegemund:
Comments:
2. *l*ist of address blocks
3. Gets or sets the transport potocol for the external endpoint: better not say 'external' endpoint
4. Gets or sets PrivateIP allocation method (Static/Dynamic). Possible values include: 'Static', 'Dynamic': private IP allocation method. Also, let's not mention possible values here. JSON schema-aware editor will show.
5. Gets or sets Provisioning state of the PublicIP resource Updating/Deleting/Failed: how can customer Set this field? Also, capitalization needs to be addressed here.
6. Gets or sets Pools of backend IP addresseses: *p*ools
7. Why are etags in the schema. I think customer does not have a way to set these.
8. Why are provisioning state in the schema. I think customer does not have a way to set these.
9. Gets or sets the Internal DNS name: *i*nternal DNS name label
10. Gets or sets DNS Settings in  NetworkInterface: extra space. Capitalization.
11. Mac address is read-only field right now. Why does it appear?
12. Gets whether this is a primary NIC on a virtual machine: that is correct. In this context, it only gets. So, why is it included in the template schema?
13. Gets or sets Security rules of network security group: this capitalization issue appears very often here at *S*ecurity. Can we fix this everywhere it occurs?

daschult:
I agree with @FrankSiegemund that it doesn't make sense for the customer to set the provisioning state of any resource. In your swagger doc, you should add a "readOnly": true property to these properties. That will trigger the generator to not include them in your schema. I'll remove them from this pull request for now though.

tools/ResourceMetaSchema.json
daschult: I wrote this ResourceMetaSchema.json file, so I think it's okay if I change it here. The description isn't really necessary at this point.

daschult:
I've fixed some of @FrankSiegemund's comments, but I don't think I've fixed them all. As I said earlier, many of the problems are because you don't have a "readonly" property in your swagger doc on properties that are read-only.

daschult:
Who are we still waiting for on this pull request?

wenwu449:
Network team is still reviewing some part of the schema.

philon-msft:
@wenwu449, can this be merged now?

PR #165, opened 2016-05-31
Add DataLakeStore test

daschult:


tests/2015-10-01-preview/Microsoft.DataLakeStore.tests.json
```diff
@@ -48,5 +48,19 @@
         }]
       }
     }
+  },
+  {
+    "name": "accounts - From Azure QuickStart templates",
+    "expectedErrors": [],
```
TianoMS: I ran this case and seems it's failing. Do we expect errors here?

daschult: There shouldn't have been any errors. Thanks for catching this.

TianoMS:
Looks good now.

PR #166, opened 2016-06-01
Version 2016-05-15 of DevTestLabs

iouri-s:


TianoMS:
@iouri-s Please add the new resource types to schemas\2014-04-01-preview\deploymentTemplate.json and schemas\2015-01-01\deploymentTemplate.json.

TianoMS:
Looks good now.

PR #177, opened 2016-06-14
Add Microsoft.RecoveryServices API schema

nilaydshah:
Microsoft.RecoveryServices API schema and tests for vaults resource type.

### DIFF
--- a/schemas/2016-06-01/Microsoft.RecoveryServices.json
+++ b/schemas/2016-06-01/Microsoft.RecoveryServices.json
@@ -1,0 +1,124 @@
+{
+  "id": "http://schema.management.azure.com/schemas/2016-06-01/Microsoft.RecoveryServices.json#",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Microsoft.RecoveryServices",
+  "description": "Microsoft Recovery Services Resource Types",
+  "resourceDefinitions": {
+    "vaults": {
+      "type": "object",
+      "properties": {
+        "type": {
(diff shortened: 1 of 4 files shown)


schemas/2016-06-01/Microsoft.RecoveryServices.json
```diff
@@ -0,0 +1,124 @@
...
+              ]
+            },
+            {
+              "$ref": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression"
+            }
+          ],
+          "description": "Gets or sets the sku name. Required for vault creation, optional for update. Possible values include: 'RS0'"
+        }
+      },
+      "tier": {
```
wenwu449: is this at the correct level? is should be under "properties", right?

nilaydshah: This is at correct level for our RP. We take sku at top level (outside the properties bag) and both name and tier are properties of sku.
If you see Microsoft.Storage is also following same - sku outside properties in 2016-01-01/Microsoft.Storage.json

wenwu449: if "tier" is a property of "sku", it should be under "properties", not the same level, the "properties" in ln82 is a schema object property, not name of property.

nilaydshah: @wenwu449 thanks for pointing out. Updated PR with the fix.

PR #182, opened 2016-06-16
Add Media Services api version 2015-10-01

BrianBlum:
Resource added to 
schemas\2014-04-01\preview\deploymentTemplate.json,
schemas\2015-01-01\deplomentTemplate.json

From tools, ran
Node validateSchemas.js
Node runSchemaTests.js

### DIFF
--- a/schemas/2015-10-01/Microsoft.Media.json
+++ b/schemas/2015-10-01/Microsoft.Media.json
@@ -1,0 +1,61 @@
+{
+  "id": "http://schema.management.azure.com/schemas/2015-10-01/microsoft.media.json#",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Microsoft.Media",
+  "description": "Microsoft Media Resource Types",
+  "resourceDefinitions": {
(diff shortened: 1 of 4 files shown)


schemas/2015-10-01/Microsoft.Media.json
```diff
@@ -0,0 +1,61 @@
...
+                    {},
+                    {
+                      "$ref": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression"
+                    }
+                  ],
+                  "description": "Microsoft.Media/mediaServices apiEndpoints "
+                },
+                "storageAccounts": {
+                  "oneOf": [
+                    {},
```
wenwu449: does this mean any json object will work?

BrianBlum: There’s a definition for an api endpoint, but it didn’t make this doc for some reason. Thanks for catching this. I used AutoRest to generate the doc, but didn’t look closely at the output.

You can close the pull request and I’ll fix the schema.

PR #188, opened 2016-06-22
Add Azure Media Service 2015-10-01 schema

BrianBlum:
Validated through VS, per test doc.

### DIFF
--- a/schemas/2015-10-01/Microsoft.Media.json
+++ b/schemas/2015-10-01/Microsoft.Media.json
@@ -1,0 +1,84 @@
+{
+  "id": "http://schema.management.azure.com/schemas/2015-10-01/microsoft.media.json#",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Microsoft.Media",
+  "description": "Microsoft Media Service Resource Type",
+  "resourceDefinitions": {
+    "mediaServices": {
+      "type": "object",
+      "properties": {
+        "type": {
+          "enum": [
+            "Microsoft.Media/mediaServices"
+          ]
+        },
+        "apiVersion": {
+          "enum": [
+            "2015-10-01"
+          ]
+        },
+        "name": {
+          "oneOf": [
+            {
+              "type": "string",
+              "pattern": "^[a-z0-9]{3,26}$"
+            },
+            {
+              "$ref": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression"
+            }
+          ]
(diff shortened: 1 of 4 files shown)


schemas/2015-10-01/Microsoft.Media.json
```diff
@@ -0,0 +1,84 @@
...
+          "items": {
+            "$ref": "#/definitions/storageAccount"
+          }
+        }
+      },
+      "type": "object"
+    },
+    "storageAccount": {
+      "properties": {
+        "id": {
```
wenwu449: please support expression for the "id" property, usually a string type property already support expression, but if there is pattern on it, we need to define $ref expression for it.

schemas/2015-10-01/Microsoft.Media.json
```diff
@@ -0,0 +1,84 @@
...
+        "apiVersion",
+        "location"
+      ],
+      "description": "Microsoft.Media/mediaServices"
+    }
+  },
+  "definitions" : {
+    "storageAccounts": {
+      "properties": {
+        "storageAccounts": {
```
wenwu449: would be good if we can support expression for the "storageAccounts" property.

BrianBlum: OK. Will add.

schemas/2015-10-01/Microsoft.Media.json
BrianBlum: Will add expression.

BrianBlum: What’s the preferred way to push this change, update just the schema, or update all 4 files?

From: Brian Blum
Sent: Wednesday, June 22, 2016 11:56 AM
To: 'Azure/azure-resource-manager-schemas' <reply@reply.github.com>; Azure/azure-resource-manager-schemas <azure-resource-manager-schemas@noreply.github.com>
Cc: Author <author@noreply.github.com>
Subject: RE: [Azure/azure-resource-manager-schemas] Add Azure Media Service 2015-10-01 schema (#188)

Will add expression.

wenwu449: Only the schema should be good. You can push the change to the BrianBlum\master branch and update this PR.

BrianBlum: I pushed my change, and it looks like the pull request now lists 2 commits. Is there an explicit action I need to take?

BrianBlum:
Thanks Tian

PR #207, opened 2016-07-21
Changed Microsoft.Compute schema

rkelsey:
I updated the Microsoft.Compute schema with three supported extensions, IaaSDiagnostics, IaaSAntimalware, and CustomScriptExtension. I also updated the tests to test support for these extensions.

### DIFF
--- a/schemas/2015-08-01/Microsoft.Compute.json
+++ b/schemas/2015-08-01/Microsoft.Compute.json
@@ -189,244 +189,10 @@
         },
         "properties": {
           "anyOf": [
-            {
-              "type": "object",
-              "properties": {
-                "publisher": {
-                  "type": "string",
-                  "minLength": 1,
-                  "description": "Microsoft.Compute/extensions - Publisher"
-                },
-                "type": {
-                  "type": "string",
-                  "minLength": 1,
-                  "description": "Microsoft.Compute/extensions - Type"
-                },
-                "typeHandlerVersion": {
-                  "type": "string",
-                  "minLength": 1,
-                  "description": "Microsoft.Compute/extensions - Type handler version"
-                },
-                "settings": {
-                  "oneOf": [
-                    { "type": "object" },
-                    { "$ref": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression" }
-                  ],
-                  "description": "Microsoft.Compute/extensions - Settings"
-                }
-              },
-              "required": [
-                "publisher",
-                "type",
-                "typeHandlerVersion",
-                "settings"
-              ]
-            },
-            {
-              "type": "object",
-              "properties": {
-                "publisher": {
-                  "enum": [
-                    "Microsoft.Azure.Diagnostics"
-                  ]
-                },
-                "type": {
-                  "enum": [
-                    "IaaSDiagnostics"
-                  ]
-                },
-                "typeHandlerVersion": {
-                  "type": "string",
-                  "minLength": 1
-                },
-                "autoUpgradeMinorVersion": {
-                  "type": "boolean"
-                },
-                "settings": {
-                  "type": "object",
-                  "properties": {
-                    "xmlCfg": {
-                      "type": "string"
-                    },
-                    "StorageAccount": {
-                      "type": "string"
-                    }
-                  },
-                  "required": [
-                    "xmlCfg",
-                    "StorageAccount"
-                  ]
-                },
-                "protectedSettings": {
-                  "type": "object",
-                  "properties": {
-                    "storageAccountName": {
-                      "type": "string"
-                    },
-                    "storageAccountKey": {
-                      "type": "string"
-                    },
-                    "storageAccountEndPoint": {
-                      "type": "string"
-                    }
-                  },
-                  "required": [
-                    "storageAccountName",
-                    "storageAccountKey",
-                    "storageAccountEndPoint"
-                  ]
-                }
-              },
-              "required": [
-                "publisher",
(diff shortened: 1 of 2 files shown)


schemas/2015-08-01/Microsoft.Compute.json
```diff
@@ -174,60 +174,241 @@
...
-        }
-      },
-      "required": [
-        "type",
-        "apiVersion",
-        "properties"
-      ],
-      "description": "Microsoft.Compute/extensions"
-    },
+		"type": "object",
```
TianoMS: Please use two spaces as indent.

tests/2015-08-01/Microsoft.Compute.tests.json
```diff
@@ -92,16 +92,76 @@
...
+        "settings": { }
+      }
+    },
+	
+	{
+      "name": "extensions - Test IaaSDiagnostics",
       "definition": "http://schema.management.azure.com/schemas/2015-08-01/Microsoft.Compute.json#/resourceDefinitions/extensions",
       "json": {
+		"name": "Test Name",
+		"location": "Test Location",
```
TianoMS: Same. Use two spaces as indent.

schemas/2015-08-01/Microsoft.Compute.json
```diff
@@ -187,46 +190,224 @@
...
-            "publisher": {
-              "type": "string",
-              "minLength": 1,
-              "description": "Microsoft.Compute/extensions - Publisher"
-            },
-            "type": {
-              "type": "string",
-              "minLength": 1,
-              "description": "Microsoft.Compute/extensions - Type"
+          "oneOf": [
```
TianoMS: Basically you are enabling three extensions. At the same time, you need to make sure the old ones still work. You may want to do this:
"anyOf": [
  {old ones which take "type" as any string}, 
  {
    "oneOf": [
        { new extension one },
        { new extension two },
        { new extension three }
    ]
  }
]

schemas/2015-08-01/Microsoft.Compute.json
```diff
@@ -187,46 +190,224 @@
...
+                  ]
+                }
+              },
+              "required": [
+                "publisher",
+                "type",
+                "typeHandlerVersion",
+                "autoUpgradeMinorVersion",
+                "settings"
+              ]
```
TianoMS: Seems the extensions share some of the properties. Can you check if we can create a common definition for them?

schemas/2015-08-01/Microsoft.Compute.json
```diff
@@ -187,46 +190,224 @@
...
+                        "false"
+                      ]
+                    },
+                    "ScheduledScanSettings": {
+                      "type": "object",
+                      "properties": {
+                        "isEnabled": {
+                          "enum": [
+                            "true",
+                            "false"
```
TianoMS: why is this string values of "true" and "false", instead of boolean type?

schemas/2015-08-01/Microsoft.Compute.json
```diff
@@ -176,6 +176,9 @@
     "extensions": {
       "type": "object",
       "properties": {
+        "name": {
+          "type": "string"
+        },
```
TianoMS: we don't need to add "name" definition here as it can be inherited. Same for "location".

tests/2015-08-01/Microsoft.Compute.tests.json
```diff
@@ -7,11 +7,10 @@
         "type": "Microsoft.Compute/availabilitySets",
         "apiVersion": "2015-05-01-preview",
         "location": "West Us",
-        "properties": {
-        }
+        "properties": { }
```
TianoMS: Let's keep the closing bracket in a separate line.

tests/2015-08-01/Microsoft.Compute.tests.json
```diff
@@ -90,22 +86,82 @@
         }
       }
     },
-
+    
+    {
+      "name": "extensions - Unsupported",
```
TianoMS: I suppose we should still allow the resource definitions which follow the old extension schema, right?

schemas/2015-08-01/Microsoft.Compute.json
rkelsey: If we keep the old definition, we'll end up exporting partial extensions since the settings property isn't fully defined. Since we're greenlighting three extensions and not others, it doesn't make sense to fully support three and partially support all others.

schemas/2015-08-01/Microsoft.Compute.json
rkelsey: This is just what the individual schemas specified.

schemas/2015-08-01/Microsoft.Compute.json
rkelsey: I think this is the best we can do. Since publisher, type, and settings are different for each extension we can't really combine them and still make sure each extension follows their specific publisher, type, and settings properties.

tests/2015-08-01/Microsoft.Compute.tests.json
```diff
@@ -90,18 +88,72 @@
...
+            },
+            "RealtimeProtectionEnabled": "true",
+            "ScheduledScanSettings": {
+              "isEnabled": "false",
+              "scanType": "Test Scan Type",
+              "day": "Test Day",
+              "time": "Test Time"
+            }
+          }
         }
```
TianoMS: Can we add a test for "customScriptExtension"?

PR #208, opened 2016-07-26
Adding template schema for Machine Learning Commitment Plans

ericwill-msft:
Adding commitmentPlans resource type to our template schema. Note that all properties of commitmentPlan ARM resources are read-only, so I have not included properties in the schema. Thanks!

### DIFF
--- a/schemas/2016-05-01-preview/Microsoft.MachineLearning.json
+++ b/schemas/2016-05-01-preview/Microsoft.MachineLearning.json
@@ -4,6 +4,32 @@
   "title": "Microsoft.MachineLearning",
   "description": "Microsoft MachineLearning Resource Types",
   "resourceDefinitions": {
+    "commitmentPlans" : {
+      "type": "object",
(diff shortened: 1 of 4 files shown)


schemas/2016-05-01-preview/Microsoft.MachineLearning.json
```diff
@@ -4,6 +4,32 @@
...
+            "Microsoft.MachineLearning/commitmentPlans"
+          ]
+        },
+        "apiVersion": {
+          "type": "string",
+          "enum": [
+            "2016-05-01-preview"
+          ]
+        },
+        "sku": {
```
TianoMS: We don't accept custom properties at top level. Please create a "properties" property and put "sku" there. I think our deployment service blocks this also. (if you write a template like this and deploy it, you should get an error.)

TianoMS:
Please ignore my previous comment. This PR is good to go.

PR #210, opened 2016-07-28
Fix resourceMetaSchema in test tools.

TianoMS:
1. support "anyOf" in resource properties. 
2. add CreateUIDefinition files to whitelist.

### DIFF
--- a/tools/ResourceMetaSchema.json
+++ b/tools/ResourceMetaSchema.json
@@ -204,6 +204,32 @@
         {
           "type": "object",
           "properties": {
+            "anyOf": {
+              "type": "array",
+              "items": {
+                "oneOf": [
+                  {
+                    "type": "object",
+                    "properties": {
+                      "$ref": {
+                        "type": "string"
+                      }
+                    }
+                  }
+                ]
+              }
+            },
+            "description": {
+              "$ref": "#/definitions/description"
+            }
+          },
+          "required": [
+            "anyOf"
+          ]
+        },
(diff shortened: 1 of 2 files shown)


tools/ResourceMetaSchema.json
```diff
@@ -204,6 +204,32 @@
         {
           "type": "object",
           "properties": {
+            "anyOf": {
```
daschult: Don't you guys usually encourage people to use "oneOf" instead of "anyOf"?

tools/ResourceMetaSchema.json
```diff
@@ -204,6 +204,32 @@
...
+              "type": "array",
+              "items": {
+                "oneOf": [
```
daschult: What's the point of having a "oneOf" with only one element?

tools/ResourceMetaSchema.json
TianoMS: That's right. We encourage people to use "oneOf" when possible. This time we have a special case for Microsoft.Compute. They need "anyOf" to describe their extensions.

tools/ResourceMetaSchema.json
TianoMS: I actually copied this entry from the one above. I can update it if you don't like them. :)

daschult: Nah, it's fine.

daschult:
Looks good to me.

PR #231, opened 2016-09-13
Fix type for pipeline resources for data factories

bgold09:
* Although the child resources are known publicly as "pipelines", the
    type name in the API route is "datapipelines".
  * This will fix resource creation and export via deployment template
    schemas.

### DIFF
--- a/schemas/2015-10-01/Microsoft.DataFactory.json
+++ b/schemas/2015-10-01/Microsoft.DataFactory.json
@@ -125,7 +125,7 @@
         }
       }
     },
-    "pipelines": {
+    "dataPipelines": {
       "description": "A logical grouping of Activities. that operate as a unit that together perform a task.",
       "type": "object",
       "required": [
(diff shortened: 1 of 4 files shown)


schemas/2015-10-01/Microsoft.DataFactory.json
```diff
@@ -141,7 +141,7 @@
           ]
         },
         "type": {
-          "enum": [ "Microsoft.DataFactory/datafactories/pipelines" ]
+          "enum": [ "Microsoft.DataFactory/datafactories/datapipelines" ]
```
wenwu449: type should be exactly same with the word in service's manifest, including casing.

bgold09: @wenwu449 should be fixed now, if I understood you correctly (changed line 128 to be "datapipelines")

schemas/2015-10-01/Microsoft.DataFactory.json
```diff
@@ -125,7 +125,7 @@
         }
       }
     },
-    "pipelines": {
+    "dataPipelines": {
```
wenwu449: you need to update the 2 deploymentTemplate.json files for this reference name change.

bgold09: forgot that; fixed

PR #232, opened 2016-09-15
Update ARM template for IotHub to include IP Filter rules

rajeevmv:


schemas/2016-02-03/Microsoft.Devices.json
```diff
@@ -241,6 +241,21 @@
             }
           ]
         },
+        "ipFilterRules": {
```
wenwu449: do you have apiVersion change for this new property?

schemas/2016-02-03/Microsoft.Devices.json
```diff
@@ -355,6 +370,32 @@
         }
       },
       "description": "The properties of the Storage Endpoint provided by the customer for file upload"
+    },
+    "IpFilterRule": {
+      "description": "IP filter Rule",
+      "type": "object",
+      "properties": {
+        "filterName": {
```
wenwu449: please add expression support for all properties.

schemas/2016-02-03/Microsoft.Devices.json
rajeevmv: No. This change is backward-compatible.

schemas/2016-02-03/Microsoft.Devices.json
rajeevmv: I don't understand what it means to add expression support. Can you please explain?

wenwu449: To support customer to use template expression for any properties, we need to have following reference: "$ref": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression"

rajeevmv: Done

schemas/2016-02-03/Microsoft.Devices.json
```diff
@@ -355,6 +370,53 @@
...
+            {
+              "description": "The action desired - accept or reject.",
+              "enum": [ "Accept", "Reject" ]
+            },
+            {
+              "$ref": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression" 
+            }
+          ]
+        },
+        "ipMask": {
```
wenwu449: "string" type property already support "expression" as expression itself is a string, the "oneOf" would make VS failed to match the property.
Please revert this ipMask property to last iteration.

rajeevmv: done

PR #233, opened 2016-09-19
Add UK regions to the deployment template

chidmdxx:


schemas/2015-01-01/deploymentTemplate.json
```diff
@@ -401,6 +401,8 @@
             "Canada East",
             "West Central US",
             "West US 2",
+            "UK South",
+            "UK West",
```
chidmdxx: we don't seem to keep 2014 current, it's missing Canada and India

PR #239, opened 2016-09-30
Add new Microsoft.Network schema

daschult:


daschult:
@TianoMS @wenwu449 Is anything stopping us from merging this pull request?

TianoMS:
@daschult I think we need to get the sign off from Network team on this. It's a large change and we are not sure if they support all the scenarios yet and if it's OK to expose them.

PR #241, opened 2016-10-03
Update Microsoft.Compute.json in 2015-08-01 to include Extension apiVersion which supports GET for template exports

jluk:
Adding "2016-03-30" to apiVersion enums for extensions and extensionsChild to support template export of extensions.

Received a code review from Purav with plans to roll this to Brazil for testing the export feature but would love additional insight if this could impact any other dependencies you see.


jluk:
Please ensure this change is not included when #238 is merged

PR #248, opened 2016-10-17
Update machine learning

daschult:


schemas/2016-05-01-preview/Microsoft.MachineLearning.json
```diff
@@ -383,9 +379,7 @@
           "oneOf": [
             {
               "type": "object",
-              "additionalProperties": {
-                "type": "object"
-              }
```
tudordid: So we are not loosing any information here ? Is this still going to be seen as a dictionary ?

daschult: Yes, "globalParameters" will still be a dictionary. With the previous schema, it was a dictionary of string -> JSON objects. With this new schema, its now a dictionary of string -> any JSON type.

daschult:
These changes are a result of the schema generating changing to support swagger discriminators. Swagger documents do not support oneOf expressions, so the way that they allow for the polymorphic-like behavior that oneOf expressions support is by using discriminators.
Previously the schema generator only supported discriminators that had one possible value (MachineLearning was previously the only swagger document with discriminators). This change is using the new generator that now supports multiple possible discriminator values.

daschult:
@TianoMS @wenwu449 Can we get this pull request merged?

TianoMS:
@daschult Left one comment. Please address it. Other than that the PR looks good.

TianoMS:
@daschult One comment that needs to address. Please take a look.

PR #250, opened 2016-10-19
Update schema for VMSS

DeepakRajendranMsft:


schemas/2015-08-01/Microsoft.Compute.json
```diff
@@ -802,13 +802,21 @@
         "name": {
           "type": "string"
         },
-        "subnet": {
-          "$ref": "#/definitions/id"
-        },
-        "loadBalancerBackendAddressPools": {
-          "type": "array",
-          "items": {
+		"properties" : {
```
daschult: Did you mean to add this extra layer of properties? This means that ipConfiguration went from matching this:
{
  "name": "blah",
  "subnet": { ... },
  "loadBalancerBackendAddressPools": [ ... ],
}
to
{
  "name": "blah",
  "properties": { ... }
}
In fact, I'm not even sure that this is a valid JSON schema. Do you have any tests written for this definition?

daschult: I just saw the second part of your pull request. Just for testing purposes, try adding "additionalProperties": false line just before the "name" property of the ipConfiguration definition, and then run the tests again. I think you'll find that they fail.

tests/2015-08-01/Microsoft.Compute.tests.json
```diff
@@ -242,7 +242,22 @@
...
+        "properties":{
+          "subnet" : {
+            "id": "resourceId" 
+          },
+          "loadBalancerBackendAddressPools": [
+            {
+              "id": "backendaddresspoolId"
+            }
+          ],
+          "loadBalancerInboundNatPools": [
```
daschult: Is it valid for these arrays to be empty? What about having two elements in the arrays? Can you add a test for those conditions?

daschult:
Looks good to me.

TianoMS:
Looks good to me.

DeepakRajendranMsft:
@TianoMS @daschult who can help me merge?

daschult:
I can't do it. Tian will have to do it.

PR #256, opened 2016-11-01
Adding VMSnapshot and VMSnapshotLinux Extension Schemas

jluk:


TianoMS:
@jluk Could you sync and merge with upstream/master? I don't see "2016-03-30" in your version of file.

schemas/2015-08-01/Microsoft.Compute.json
```diff
@@ -1592,6 +1596,212 @@
         "settings",
         "protectedSettings"
       ]
+    },
+    "VMSnapshot":{
```
TianoMS: The two definitions don't follow the right pattern. Please look at the other extension definitions to correct it.

jluk:
Made changes mentioned, I removed the top parent properties layer in the snapshot schemas. I don't see any other schema patterns requiring the properties that exist outside "properties" returned from GET calls. I don't think those are required for export/redeploy since they look specific to the deployment. Can you confirm that?

schemas/2015-08-01/Microsoft.Compute.json
jluk: Could you link the line of definition? I see location defined for 3 resources, but extensions is not one of them. I ask because "tags" is a property of the snapshot extension, but I can't find it defined for "extensions".

TianoMS: Synced offline. "location" and "tags" are defined in base types in "deploymentTemplate.json". Resource types can override them if needed.

PR #258, opened 2016-11-10
Search: Adding schema and tests for Azure Search ARM 2015-08-19

brjohnstmsft:
FYI @TianoMS @daschult @seansaleh

### DIFF
--- a/schemas/2015-08-19/Microsoft.Search.json
+++ b/schemas/2015-08-19/Microsoft.Search.json
@@ -39,7 +39,7 @@
               "$ref": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression"
             }
           ],
-          "description": "The geographic location of the resource. This must be one of the supported and registered Azure Geo Regions (for example, West US, East US, Southeast Asia, and so forth)."
+          "description": "The geographic location of the resource. This must be one of the registered Azure Geo Regions that supports Azure Search (for example, West US, East US, Southeast Asia, and so forth). You can determine which regions support Azure Search programmatically by obtaining information about the Azure Search Resource Provider. See https://aka.ms/search-rp-info for details."
         },
         "sku": {
           "oneOf": [
@@ -84,8 +84,7 @@
           "oneOf": [
             {
               "type": "integer",
-              "minimum": 1,
-              "maximum": 12
+              "enum": [ 1, 2, 3, 4, 6, 12 ]
             },
             {
               "$ref": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression"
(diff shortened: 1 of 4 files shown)


schemas/2015-08-19/Microsoft.Search.json
```diff
@@ -0,0 +1,139 @@
...
+            {
+              "$ref": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression"
+            }
+          ],
+          "description": "The number of replicas in the Search service. If specified, it must be a value between 1 and 12 inclusive for standard SKUs or between 1 and 3 inclusive for basic SKU."
+        },
+        "partitionCount": {
+          "oneOf": [
+            {
+              "type": "integer",
```
seansaleh: We should have this as a list, since values of 5,7,8,9,10, and 11 are invalid

schemas/2015-08-19/Microsoft.Search.json
```diff
@@ -0,0 +1,139 @@
...
+      "type": "object",
+      "properties": {
+        "name": {
+          "oneOf": [
+            {
+              "type": "string",
+              "enum": [
+                "free",
+                "basic",
+                "standard",
```
seansaleh: By hardcoding SKU's here we are committing to updating this schema as we introduce new SKU's. Do we want such a commitment?

schemas/2015-08-19/Microsoft.Search.json
```diff
@@ -0,0 +1,139 @@
...
+        "location": {
+          "oneOf": [
+            {
+              "type": "string"
+            },
+            {
+              "$ref": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression"
+            }
+          ],
+          "description": "The geographic location of the resource. This must be one of the supported and registered Azure Geo Regions (for example, West US, East US, Southeast Asia, and so forth)."
```
seansaleh: This must be one of the supported and registered Azure Geo Regions _**where Azure Search is available**_ (and maybe a link to our docs / definitive source?)

brjohnstmsft: I don't think we document that info anywhere since it changes over time, but it is possible to obtain it programmatically. I'll update the description with a link to info on how to do this.

schemas/2015-08-19/Microsoft.Search.json
brjohnstmsft: We're already committed to updating the Swagger and SDKs. I don't see this as being any different. Besides, not hard-coding the SKUs gives no clue as to what the value should be.

brjohnstmsft:
@seansaleh I've addressed your comments.

seansaleh:
:shipit:

PR #264, opened 2016-12-03
fix enum for bmcctmagentwindows

jluk:


jluk:
Could we get this merged and rolled out on this week's train?

PR #267, opened 2016-12-12
Add Schema for Container Registry

matthieuklotz:
This PR will add schema for Azure Container Registry that currently is in preview.

PR #270, opened 2017-01-11
Fix incorrect $ref in pull request #267 (ContainerRegistry)

xinsnake:
The newly changed schema in #267 contains an incorrect reference to Microsoft.ContainerRegistry.json. It caused the schema validation to fail.

P.S. It took me 3 hours to work out why my schema validation doesn't work.


TianoMS:
@xinsnake Thanks for the fix. Just curious, the char "/" should be optional. What breaks for you?

xinsnake:
@TianoMS thanks for your comment.

I was trying to use Newtonsoft.Json.Schema to validate an ARM template against the 2015-01-01 schema. The validation failed with unable to resolve "http://schema.management.azure.com/schemas/2016-06-27-preview/Microsoft.ContainerRegistry.json#resourceDefinitions/registries". I tried to use a version of the schema with the slash "/", and it worked.

I haven't had a detailed look into the JSON schema, and it may be an issue with Newtonsoft.Json.Schema, which I'll investigate and submit a pull request for that if it's a bug in their system.

For the moment I think it's better to have the slash added.

Thanks!

TianoMS:
@xinsnake  I did a bit more investigation on this. You are correct. The JSON schema requires the slash "/". Thanks for the fix.

PR #274, opened 2017-01-19
properties is not a required field for Microsoft.Storage/storageAccounts

ramach-msft:
Starting in 2016-01-01, properties is used to specify some more advanced use cases.
The common scenario is to deploy with no properties field at all.

If the user doesn't want to see schema problems, he needs to insert an empty properties field
( "properties": { } ) in his resource declaration.
This PR will improve the experience.

### DIFF
--- a/schemas/2016-01-01/Microsoft.Storage.json
+++ b/schemas/2016-01-01/Microsoft.Storage.json
@@ -79,8 +79,7 @@
         "apiVersion",
         "sku",
         "kind",
-        "location",
-        "properties"
+        "location"
       ],
       "description": "Microsoft.Storage/storageAccounts"
     }


TianoMS:
@ramach-msft I tried to deploy a template with one storage account which doesn't have properties. The deployment was rejected by Storage during preflight.

ramach-msft:
@TianoMS Are you positive you used the 2016-01-01 apiVersion ?

This is a valid deployment with a template that does not contain "properties" in the StorageAccount:

```javascript
{
  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
  "contentVersion": "1.0.0.0",
  "resources": [
    {
      "name": "[concat('sta', uniqueString(resourceGroup().id))]",
      "type": "Microsoft.Storage/storageAccounts",
      "location": "[resourceGroup().location]",
      "apiVersion": "2016-01-01",
      "sku": {
        "name": "Standard_LRS"
      },
      "dependsOn": [ ],
      "tags": {
        "displayName": "sta"
      },
      "kind": "Storage"
    }
  ]
}
```

TianoMS:
@ramach-msft double checked on this. It works. Do we need anyone from MS.Storage to sign off this?

ramach-msft:
@TianoMS It's probably the best idea, since they own the schema. Do you have a contact to ask for the signoff ?

PR #283, opened 2017-02-22
Update Microsoft.Compute.json

QuasarSE:
The VM Extension settings key "storageAccount" was misspelled as "StorageAccount", which shows up as intellisense errors.  Corrected.

### DIFF
--- a/schemas/2015-08-01/Microsoft.Compute.json
+++ b/schemas/2015-08-01/Microsoft.Compute.json
@@ -1024,13 +1024,13 @@
             "xmlCfg": {
               "type": "string"
             },
-            "StorageAccount": {
+            "storageAccount": {
               "type": "string"
             }
           },
           "required": [
             "xmlCfg",
-            "StorageAccount"
+            "storageAccount"
           ]
         },
         "protectedSettings": {


schemas/2015-08-01/Microsoft.Compute.json
```diff
@@ -1024,13 +1024,13 @@
             "xmlCfg": {
               "type": "string"
             },
-            "StorageAccount": {
+            "storageAccount": {
```
TianoMS: can we update the test files in tests/2015-08-01/Microsoft.Compute.tests.json?

TianoMS:
@jluk can you take a look at this change?

jluk:
LGTM

TianoMS:
@QuasarSE there's one comment. Can you please address it?

QuasarSE:
@TianoMS - do you mean the LGTM?  i.e., @jluk is fine with the change?  so we're good to go & you'll merge?

jluk:
@QuasarSE I think Tian is referring to your test file in tests/2015-08-01/Microsoft.Compute.tests.json

QuasarSE:
Thx @jluk.   I don't know how to add another file to this PR as the original branch seems missing :-(   I was creating a new PR with this file + the test file and realized that this will be a breaking change -- unless ARM reads properties as case-independent -- e.g. storageAccount == StorageAccount.     We really should fix this one for consistency, but it's not worth breaking existing templates -- so would it (break them)?   Thx.

TianoMS:
@QuasarSE Please check this PR: https://github.com/QuasarSE/azure-resource-manager-schemas/pull/2.

PR #284, opened 2017-03-02
add missing type for virtualMachineScaleSetExtension

hoffmann:


hoffmann:
add missing type for virtualMachineScaleSetExtension

schemas/2015-08-01/Microsoft.Compute.json
```diff
@@ -855,7 +855,8 @@
             "typeHandlerVersion"
           ]
         }
-      }
+      },
+      "type": "object"
```
TianoMS: Is this breaking anything?

PR #288, opened 2017-03-08
Fix Typo  in dockerExtension

hoffmann:


TianoMS:
@hoffmann Thanks for the fix. Can you also make the change for 2015-08-01/Microsoft.Compute.json?

hoffmann:
done

PR #290, opened 2017-03-14
Added $schema Property to all versions of CreateUIDefinition schema

QuasarSE:
Necessary to include "$schema" in a JSON UI source file to have VS Code and other static validators associate the correct schema.   Left $schema as an optional property so as not to break existing UI definitions.


TianoMS:
@yuyez can you take a look at this change?

PR #294, opened 2017-03-23
UI: VM SizeSelector - min# recommendedSizes 3->1

QuasarSE:
Decrease the minimum recommendedSizes list size from 3 to 1 on the CustomUI SizeSelector control.


QuasarSE:
Changes across the four UI versions to decrease the minimum size of the recommendedSizes list of the VM Size selector control.   This is to accommodate solutions where even the current 3 is too many.

QuasarSE:
This one is superseded by #296.  i.e., they don't conflict, but #296 contains this set of changes.

PR #296, opened 2017-03-24
Combined recommendedSizes and regex fixes

QuasarSE:
Combination pull request with both of:
1. Change for VM SizeSelector Control to reduce the minimum size of the recommendedSizes array from 3 to 1.
2. Expose the regex and validationMessage constraints on the UserNameTextBox control.

### DIFF
--- a/schemas/0.0.1-preview/CreateUIDefinition.CommonControl.json
+++ b/schemas/0.0.1-preview/CreateUIDefinition.CommonControl.json
@@ -459,7 +459,13 @@
           "properties": {
             "required": {
               "type": [ "boolean", "string" ]
-            }
+            },
+            "regex": {
+              "type": "string"
+            },
+            "validationMessage": {
+              "type": "string"
+            }            
           },
           "additionalProperties": false
         },
--- a/schemas/0.0.1-preview/CreateUIDefinition.CommonControl.json
+++ b/schemas/0.0.1-preview/CreateUIDefinition.CommonControl.json
@@ -460,7 +460,7 @@
             "required": {
               "type": [ "boolean", "string" ]
             },
-            "regex": {
+            "regex": {
               "type": "string"
             },
             "validationMessage": {
--- a/schemas/0.0.1-preview/CreateUIDefinition.ProviderControl.json
+++ b/schemas/0.0.1-preview/CreateUIDefinition.ProviderControl.json
@@ -170,7 +170,7 @@
           "items": {
             "type": "string"
           },
-          "minItems": 3
+          "minItems": 1
         },
         "constraints": {
           "oneOf": [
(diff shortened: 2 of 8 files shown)


QuasarSE:
Note that this pr is a superset of #294.   If this one is applied, no need to apply #294.

TianoMS:
@yuyez can you review this PR?

yuyez:
@tabrezm Could you please take a look? The suggested changes differ from the element docs.

TianoMS:
@tabrezm could you review this change?

tabrezm:
@yuyez I'm OK updating the docs to match, as long as this behavior is functionally supported by CreateUiDef. Can you confirm?

schemas/0.0.1-preview/CreateUIDefinition.ProviderControl.json
```diff
@@ -170,7 +170,7 @@
           "items": {
             "type": "string"
           },
-          "minItems": 3
+          "minItems": 1
```
tabrezm: Could we make it 0? Recommendations shouldn't be mandatory.

schemas/0.0.1-preview/CreateUIDefinition.CommonControl.json
```diff
@@ -459,7 +459,13 @@
           "properties": {
             "required": {
               "type": [ "boolean", "string" ]
-            }
+            },
```
tabrezm: Is this actually supported properly for other elements? I"m worried that there may be at least one element where this behavior won't work as expected if we push this schema change.

yuyez: This schema change only affects UserNameTextBox which I think supports regex and validationMessage.

schemas/0.0.1-preview/CreateUIDefinition.ProviderControl.json
yuyez: I saw our code checking for at least one.

schemas/0.0.1-preview/CreateUIDefinition.CommonControl.json
tabrezm: Oops, sorry. Missed this when scrolling up in the PR. LGTM

schemas/0.0.1-preview/CreateUIDefinition.ProviderControl.json
tabrezm: OK, let's bring this down to 1 in the schema; we can do a code change to support 0 if users request it.

tabrezm:
@TianoMS feel free to merge this. I'll update the wiki once this PR is live.

PR #307, opened 2017-04-11
Regen SQL schemas from swagger.

jaredmoo:
azure-rest-api-specs commit 06979eee90715e92d157418035d56510aeb93782


TianoMS:
@jaredmoo Please add test cases for your schemas.

ravbhatnagar:
@jaredmoo - Just wanted to confirm if you are planning on adding the tests for this. We should have the updated numbers before next months review so that we can show progress.

PR #326, opened 2017-06-23
Add sql data masking policy and rules, and server administrators

jaredmoo:
There are also some minor changes to other resource types due to regenerating from latest swagger.

### DIFF
--- a/schemas/2014-04-01/Microsoft.Sql.json
+++ b/schemas/2014-04-01/Microsoft.Sql.json
@@ -66,10 +66,10 @@
                 "$ref": "#/definitions/servers_firewallRules_childResource"
               },
               {
-                "$ref": "#/definitions/servers_databases_childResource"
+                "$ref": "#/definitions/servers_backupLongTermRetentionVaults_childResource"
               },
               {
-                "$ref": "#/definitions/servers_backupLongTermRetentionVaults_childResource"
+                "$ref": "#/definitions/servers_databases_childResource"
               }
             ]
           }
@@ -218,6 +218,9 @@
                 "$ref": "#/definitions/servers_databases_geoBackupPolicies_childResource"
               },
               {
+                "$ref": "#/definitions/servers_databases_dataMaskingPolicies_childResource"
+              },
+              {
                 "$ref": "#/definitions/servers_databases_securityAlertPolicies_childResource"
               },
               {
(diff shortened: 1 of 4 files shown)


schemas/2014-04-01/Microsoft.Sql.json
```diff
@@ -392,7 +492,6 @@
         }
       },
       "required": [
-        "name",
```
TianoMS: >"name", [](start = 8, length = 7)

Is "name" not required?

schemas/2014-04-01/Microsoft.Sql.json
```diff
@@ -991,7 +1200,6 @@
           "description": "The administrator login password (required for server creation)."
         }
       },
-      "required": [ "administratorLogin", "administratorLoginPassword" ],
```
TianoMS: >"required": [ "administratorLogin", "administratorLoginPassword" ], [](start = 6, length = 67)

Is this by design?

tests/2014-04-01/Microsoft.Sql.tests.json
```diff
@@ -405,29 +501,24 @@
...
         },
         {
-          "message":"Data does not match any schemas from \"oneOf\"",
-          "dataPath":"/properties",
-          "schemaPath":"/properties/properties/oneOf",
-          "subErrors":[
+          "message": "Data does not match any schemas from \"oneOf\"",
+          "dataPath": "/properties",
+          "schemaPath": "/properties/properties/oneOf",
+          "subErrors":     [
```
TianoMS: >      [](start = 22, length = 5)

format?

schemas/2014-04-01/Microsoft.Sql.json
jaredmoo: These are required for creating a server, but not for updating. I made this required earlier so that it's clear they are required for create, but now I understand that update must also be supported so they should not be required.

schemas/2014-04-01/Microsoft.Sql.json
jaredmoo: Fixed.

tests/2014-04-01/Microsoft.Sql.tests.json
jaredmoo: Done

schemas/2014-04-01/Microsoft.Sql.json
```diff
@@ -84,6 +87,55 @@
...
     },
+    "servers_administrators": {
+      "type": "object",
+      "properties": {
+        "name": {
+          "oneOf": [
+            {
+              "type": "string",
+              "enum": [
+                "activeDirectory"
```
TianoMS: >activeDirectory [](start = 17, length = 15)

Is this correct? I don't think ARM accepts this name. For nested resources, the name needs to be something like "serverName/adminName".

schemas/2014-04-01/Microsoft.Sql.json
```diff
@@ -274,6 +329,103 @@
...
     },
+    "servers_databases_dataMaskingPolicies": {
+      "type": "object",
+      "properties": {
+        "name": {
+          "oneOf": [
+            {
+              "type": "string",
+              "enum": [
+                "Default"
```
TianoMS: >Default [](start = 17, length = 7)

same here.

schemas/2014-04-01/Microsoft.Sql.json
jaredmoo: Yes this is the only valid name. @fearthecowboy sounds like autorest should not be generating this

jaredmoo: Removed enum.

schemas/2014-04-01/Microsoft.Sql.json
jaredmoo: Removed enum.

schemas/2014-04-01/Microsoft.Sql.json
```diff
@@ -84,6 +87,52 @@
...
+      "type": "object",
+      "properties": {
+        "name": {
+          "oneOf": [
+            {
+              "type": "string"
+            },
+            {
+              "$ref": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression"
+            }
```
TianoMS: If it's just string, you don't need the option for expression. You can remove the "oneOf".

TianoMS: Same for the others. 

---
In reply to: [124428113](https://github.com/Azure/azure-resource-manager-schemas/pull/326#discussion_r124428113) [](ancestors = 124428113)

jaredmoo: This is generated by autorest, I don't want to make manual unnecessary edits here. Can it stay as is?

jaredmoo: @fearthecowboy

fearthecowboy: It shouldn't hurt anything. Weird tho' ... I'm not sure why it'd generate that. 

I filed a bug, but I can't fix it right yet... https://github.com/Azure/autorest/issues/2401

jaredmoo: Thanks, @TianoMS can you merge as-is and I will file a separate PR to regenerate when Azure/autorest#2401 is fixed?

PR #331, opened 2017-07-18
Update to correct ID API version for schema

QuasarSE:
Corrected id version to match the path for the file.  Note that deploymentTemplate.json directly references both this version and the 2016-09-01 schema files.


TianoMS:
Good catch.

PR #336, opened 2017-07-26
Regen sql from latest swagger.

jaredmoo:
Swagger commit id is 723d7ee6b699121d01d2f71ccefcc7b2183e6361


jaredmoo:
@TianoMS Is the markdown publish bug fixed?

TianoMS:
@jaredmoo are you talking about the issue in publishing .md files? That issue is not fixed yet...

TianoMS:
there are failed test cases.

jaredmoo:
Ok, opened https://github.com/Azure/azure-resource-manager-schemas/issues/338 for the publishing issue. Fixed tests.

schemas/2014-04-01/Microsoft.Sql.json
```diff
@@ -1539,6 +1697,52 @@
...
+      "type": "object",
+      "properties": {
+        "name": {
+          "oneOf": [
+            {
+              "type": "string"
+            },
+            {
+              "$ref": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression"
+            }
```
TianoMS: you don't need this since "name" is a string.

jaredmoo: Understood, but it's generated by AutoRest. Does it make a material difference to customer experience? If yes, we should file a bug on autorest. If no, let's leave it.

TianoMS: It should not cause any behavioral difference. I'm OK with keeping it.

PR #348, opened 2017-08-18
Update policyDefinition rule schema

pilor:
The original policyDefinition rule schema lives in schemas/2015-10-01-preview.  This update brings it in line with the currently available rule language and adds tests.

See https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-policy for details on Azure Resource Policy.

### DIFF
--- a/schemas/2016-12-01/policyDefinition.json
+++ b/schemas/2016-12-01/policyDefinition.json
@@ -1,0 +1,269 @@
+{
+  "id": "http://schema.management.azure.com/schemas/2016-12-01/policyDefinition.json#",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Policy Definition",
+  "description": "This schema defines Azure resource policy definition, please see https://azure.microsoft.com/en-us/documentation/articles/resource-manager-policy/ for more details.",
+  "type": "object",
+  "properties": {
+    "if": {
+      "oneOf": [
+        { "$ref": "#/definitions/condition" },
+        { "$ref": "#/definitions/operatorNot" },
+        { "$ref": "#/definitions/operatorAnyOf" },
+        { "$ref": "#/definitions/operatorAllOf" }
+      ]
+    },
+    "then": {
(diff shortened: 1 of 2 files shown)


pilor:
@cyl3392207 Updates to policyDefinitino schema

pilor:
@TianoMS, are you the right person to review these changes?  This is updating the initial policyDefinition schema that was published in 2015-10-01-preview.

TianoMS:
@pilor Yes, I'll take a look.

TianoMS:
@pilor the format of the new schema file looks like different from the other schemas we have. Is it by design? Also, do you need to expose it resource types in deploymentTemplate.json?

kiazhi:
Hi @TianoMS & @pilor,
Since I'm working on policyDefinition recently and has done some digging around to find this Pull Request #348, may I know when will this [2016-12-01/policyDefinition.json](http://schema.management.azure.com/schemas/2016-12-01/policyDefinition.json#) be live?

My reason for asking is that the current [2015-10-01-preview/policyDefinition.json](http://schema.management.azure.com/schemas/2015-10-01-preview/policyDefinition.json#) doesn't comply to the [Resource policy overview](https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-policy) documentation where I cannot use `"match":"####-???-##"` within the `"allOf": [...]` or `"anyOf": [...] `

pilor:
@TianoMS, this is a bit different than the other schemas.  It isn't intended for template use, it is purely the JSON schema of Azure Resource Policy rules (see [docs](https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-policy)).  The existing version is in 2015-10-01-preview, this is an update to that.  However it got through back then should still be relevant now.

@kiazhi, I'm not sure how long it takes to be published on schema.management.azure.com once the PR is merged.  However, you can always take the updated schema directly out of git and use it locally temporarily if you are blocked.

PR #358, opened 2017-09-28
added true option to supportsHttpTrafficOnly property

jimoneil:


schemas/2017-06-01/Microsoft.Storage.json
```diff
@@ -464,7 +464,8 @@
             {
               "type": "boolean",
               "enum": [
-                "false"
+                "false",
+				"true"
```
TianoMS: >				 [](start = 0, length = 4)

please use spaces instead of tabs.

JasonYang-MSFT: @TianoMS Is this issue cause by autorest not generating boolean type correctly? The source file is using boolean and didn't specify the values https://github.com/Azure/azure-rest-api-specs/blob/current/specification/storage/resource-manager/Microsoft.Storage/2017-06-01/storage.json#L1143

PR #361, opened 2017-10-02
Add 09-01 templates

DeepakRajendranMsft:
@TianoMS please take a look


Tiano2017:
could you add test cases for the new schema?

DeepakRajendranMsft:
@Tiano2017 no, we dont plan to add tests, it cannot scale for us to add tests here every time we instroduce an api version. we are using the autorest code generator. why should we test the generator. @salameer as FYI

DeepakRajendranMsft:
@Tiano2017  this has been opened for a really long time. can we merge this please

Tiano2017:
@DeepakRajendranMsft test cases are required. We had different discussions with partner teams but we decided that we require at least a few basic test cases. It happened in the past that AutoRest has bugs or breaking changes. We don't want those issues to get into this repo and break schema scenarios.

DeepakRajendranMsft:
why are there generic tests to catch AutoRest regressions? why should every service team write the same form of tests?

Tiano2017:
They are not designed to catch AutoRest regressions, but it can catch some issues that AutoRest introduces.  AutoRest releases new versions every month or so. The new schemas generated may contain changes made by AutoRest, which might not be valid for schemas. 

Not quite get your second question. Ping me on email or Skype for further discussion.

arbitertl:
2017-09-01 JSON Schema for Microsoft.Network is missing

arbitertl:
Application Gateways are not referenced in deploymentTemplate.json files

PR #364, opened 2017-10-11
Add support for ApiSchemas and Certificates

solankisamir:
- Add the Schema resource as child of Apis resource
- Add properties for defaultSslBinding and Custom Certificates in Schemas


solankisamir:
@Tiano2017 for your reference

schemas/2017-03-01/Microsoft.ApiManagement.json
```diff
@@ -346,6 +349,53 @@
       ],
       "description": "Microsoft.ApiManagement/service/apis/policies"
     },
+    "service_apis_schemas": {
```
TianoMS: >service_apis_schemas [](start = 5, length = 20)

do you need to add this to deploymentTemplate.json?

solankisamir: #fixed, added test for resource

PR #370, opened 2017-10-25
[ACR] Add acr schema for all api-versions

djyou:
Add or update `Microsoft.ContainerRegistry` schema for the following api-version:
```
2016-06-27-preview
2017-03-01
2017-06-01-preview
2017-10-01
```


SajayAntony:
LGTM

Tiano2017:
LGTM

schemas/2016-06-27-preview/Microsoft.ContainerRegistry.json
```diff
@@ -34,60 +39,82 @@
...
-          "type": "object",
-          "properties": {
-            "name": {
-              "type": "string"
+        "adminUserEnabled": {
+          "oneOf": [
+            {
+              "type": "boolean",
+              "enum": [
+                "false"
```
TianoMS: >"false" [](start = 16, length = 7)

Is this correct? This is a string value, while the type says "boolean".

TianoMS: same in the other api versions. 

---
In reply to: [148637443](https://github.com/Azure/azure-resource-manager-schemas/pull/370#discussion_r148637443) [](ancestors = 148637443)

djyou: This does seem odd, but is this AutoRest behavior? I did see the same issue in other schemas such as https://github.com/Azure/azure-resource-manager-schemas/blob/0f742766eb22be8d2807283e8055221d373ebf79/schemas/2017-06-01/Microsoft.Storage.json#L467
What is the workaround for this issue?

djyou: In fact, the issue is not only that `"false"` is a string, but it should also be an enum with both `true` and `false` values.

Tiano2017: Right. can you check your swagger file and see if it has the correct definition?

djyou: I suppose so.
https://github.com/Azure/azure-rest-api-specs/blob/current/specification/containerregistry/resource-manager/Microsoft.ContainerRegistry/2017-10-01/containerregistry.json#L1206
It seems that AutoRest is not treating `boolean` with default value properly.

djyou: @Tiano2017 Any update on this? Since this seems to be an AutoRest issue and it has been spread in published schema for many other services, can we have this merged and have a general cleanup later on? If not possible, we will have to request an exception of delaying the release of ACR schema as we had committed to have them a few weeks ago. Thanks. /cc @sajayantony

Tiano2017: works for me. I'll merge this PR.

djyou:
@Tiano2017 Can you please review and merge this PR as the issue has been fixed from AutoRest? Thanks.

PR #371, opened 2017-10-30
generated schemas that are not present

fearthecowboy:
I generated the schemas that are in azure/azure-rest-api-specs and created this PR that fills in the blank for files that are not in the current schemas repository.

I don't have tests, and have not done any other testing because I have no experience with the individual services.


Tiano2017:
Thanks for the work. Do we need to notify each RP team to review this? Also, our schema publishing operation has some limitation right now because github api can throttle us if we make more than 60 calls in 1 hour. So we may not be able to merge this PR till we get a solution.

PR #374, opened 2017-11-14
Add Azure Import/Export schema file

leoz-ms:
Add Azure Import/Export schema file


Tiano2017:
hello @leoz-ms, please add basic tests for your new schema. You can look at the other RP schemas for reference.

leoz-ms:
Added test file.

PR #387, opened 2018-01-03
Add DNS resources for 2017-09-01 Microsoft.Network API version

nkindberg:


nkindberg:
DNS resources are under Microsoft.Network RP but in a different rest API spec so they did not get picked-up during previous generation of schema for this API version. 

Assuming the correct thing to do is to add them to existing Microsoft.Network schema. Or should they go in a separate file?

Tiano2017:
We have a feature to support separated schema files but it's not ready yet. So you need to add them to the existing schema. BTW, the 2017-09-01 network schema still has unresolved issue about circular dependency in the definitions. I think it would be better to resolve that first.

PR #390, opened 2018-01-08
Undo adf resource schemas - there are a couple issues we need to resolve before these will work.

deata:


GamerLivingWill:
Thanks!

PR #392, opened 2018-01-11
Added 2016-10-01 and 2017-07-01 schemas for Microsoft.Logic

swylezol:


tools/package-lock.json
```diff
@@ -0,0 +1,837 @@
+{
+  "name": "template-schema-validation-tools",
```
TianoMS: >"name": "template-schema-validation-tools", [](start = 2, length = 43)

do we need this file?

swylezol: Hmm not sure why it got included. I think this is a temp file. Removed.

PR #400, opened 2018-01-24
Streamanalytics updates

thbagley:
Adding necessary resources to resourceDefinitions. Tried my best to be consistent with other schemas in the folders.


Tiano2017:
@thbagley is the change generated from swagger?

thbagley:
No, but I did my best to match it with the other schemas. Anything in particular that needs fixing?

Tiano2017:
I don't see anything specific that needs fixing. If the same change is not in  swagger, it might be overwritten later if someone generate it from swagger.

thbagley:
Our team is aware that we're managing this without swagger, so we should be alright. Thanks for double-checking.

Tiano2017:
Looks good. Will merge.

PR #421, opened 2018-03-11
Remove unwanted properties from DataMigration/services

haitsongmsft:
We will not expose public key properties and NIC as these will not be used in template deployments, and it is created during service creation behind scene. If added into template, it will not be respected during deployment


schemas/2017-11-15-privatepreview/Microsoft.DataMigration.json
```diff
@@ -775,14 +775,6 @@
...
       "type": "object",
       "properties": {
-        "publicKey": {
-          "type": "string",
-          "description": "The public key of the service, used to encrypt secrets sent to the service"
-        },
-        "virtualSubnetId": {
-          "type": "string",
-          "description": "The ID of the Microsoft.Network/virtualNetworks/subnets resource to which the service should be joined"
-        }
```
TianoMS: removing these properties leaves an empty property bag. This can be a problem because an empty property bag can match any object.

schemas/2017-11-15-privatepreview/Microsoft.DataMigration.json
```diff
@@ -775,14 +775,6 @@
...
       },
       "required": [
         "virtualSubnetId"
```
TianoMS: >virtualSubnetId [](start = 9, length = 15)

property is still in required list.

haitsongmsft: Fixed, putting the pubkey back as placeholder

schemas/2017-11-15-privatepreview/Microsoft.DataMigration.json
haitsongmsft: putting the pubkey back as placeholder

haitsongmsft:
let me make another change, we don't want to give publickey in response

Tiano2017:
Looks like all the changes are reverted. do you still need this PR?

haitsongmsft:
Yes, I changed to remove public key only, and leave subnet alone

haitsongmsft:
please still CR this, we need to remove public key.

.gitignore
```diff
@@ -1,3 +1,4 @@
 .vscode
 js-coverage
-node_modules
\ No newline at end of file
+node_modules
+/tools
```
TianoMS: we can't ignore everything in tools.

haitsongmsft: removed

.gitignore
```diff
@@ -1,3 +1,3 @@
 .vscode
 js-coverage
-node_modules
\ No newline at end of file
+node_modules
```
Tiano2017: Looks like there are some white space changes? Can you revert the changes?

haitsongmsft: done

PR #422, opened 2018-03-14
Updated Traffic Manager schemas

EvgenyAgafonchikov:


schemas/2017-05-01/Microsoft.Network.json
```diff
@@ -4,306 +4,362 @@
   "title": "Microsoft.Network",
   "description": "Microsoft Network Resource Types",
   "resourceDefinitions": {
-    "trafficManagerProfiles": {
```
TianoMS: >trafficManagerProfiles [](start = 5, length = 22)

why is this changed? I believe the current resource type has the recommended casing.

EvgenyAgafonchikov: All the stuff was autogenerated basing on current Traffic Manager state from azure-rest-api-specs. It looks like all the entries have lower case: [sample](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/trafficmanager/resource-manager/Microsoft.Network/stable/2017-05-01/trafficmanager.json#L18) which is different from [NSG](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/network/resource-manager/Microsoft.Network/stable/2018-01-01/networkSecurityGroup.json#L39) for example.

@dilukose, @KumudD, could you please have a look? If this change is not expected, should we fix schemas manually?

MikhailTryakhov: @TianoMS can you please help to resurrect discussion?

EvgenyAgafonchikov:
closing as outdated

PR #424, opened 2018-03-15
adding new 2018-01-22 api version for iot hubs

kvish:


kvish:
@akumardu

schemas/2018-01-22/Microsoft.Devices.json
```diff
@@ -0,0 +1,634 @@
+{
+  "id": "https://schema.management.azure.com/schemas/2016-02-03/Microsoft.Devices.json#",
```
TianoMS: >2016-02-03 [](start = 53, length = 10)

this should be 2018-01-22.

PR #427, opened 2018-03-19
Remove format "uuid" since it's not supported by Visual Studio.

Tiano2017:


Tiano2017:
@vrdmr can you take a look at this PR?

Tiano2017:
Yes, I think so. could you send a PR for that?

PR #429, opened 2018-03-20
Updated schema documents

akumardu:
Updated schema documents from latest


Tiano2017:
@akumardu  do you need to update 2015-01-01/deploymentTemplate.json and 2014-01-01-preview/deploymentTemplate.json? I see you have new resource types. Also please update your test for the change.

PR #439, opened 2018-04-11
Include Integration Account Batch Configurations to Microsoft.Logic

rarayudu:


schemas/2016-06-01/Microsoft.Logic.json
```diff
@@ -485,6 +485,66 @@
...
+          "description": "The integration account batch configuration properties."
+        }
+      },
+      "required": [
+        "type",
+        "apiVersion",
+        "properties"
+      ],
+      "description": "Microsoft.Logic/integrationAccounts/batchConfigurations"
     }
```
TianoMS: the change doesn't parse as valid json. can you double check?

PR #446, opened 2018-04-24
Added schema and tests for scheduledQueryRules resource in Microsoft.Insights

Adarsh1994-MS:
Added schema and tests for scheduledQueryRules resource in Microsoft.Insights


schemas/2018-04-16/Microsoft.Insights.json
```diff
@@ -0,0 +1,574 @@
...
+          "type": "string",
+          "description": "the name of the metric that defines what the rule monitors."
+        },
+        "metricResourceUri": {
+          "type": "string",
+          "description": "the resource identifier of the resource the rule monitors."
+        },
+        "timeGrain": {
+          "type": "string",
+          "format": "duration",
```
TianoMS: >"format": "duration", [](start = 10, length = 21)

duration format is not supported by Visual Studio. can we remove these?

Adarsh1994-MS:
AutoRest generated the MetricTriggerType definition. A definition with same name has been for metricAlerts in Microsoft.Insights RP and it looks like autorest picked up that definition. Used autorest again specifically for the scheduledQueryRules resource and also verified all definitions manually. Also added a new test for MetricTrigger property.

PR #456, opened 2018-05-14
Introduce Maps

jp94:


schemas/2018-05-01/Microsoft.Maps.json
```diff
@@ -0,0 +1,81 @@
...
+            },
+            {
+              "$ref": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression"
+            }
+          ],
+          "description": "The SKU of this account."
+        }
+      },
+      "required": [
+        "name",
```
TianoMS: >name [](start = 9, length = 4)

don't you have a property bag for this resource? Most resources have a property called "properties".

jp94: This is our current template that we currently accept, and is generated from our REST API specs.
We pass in an empty list for properties.

```json
{
    "$schema": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json#",
    "contentVersion": "1.0.0.0",
    "parameters": {
        "name": {
            "type": "string"
        },
        "pricingSpec": {
            "type": "string"
        }
    },
    "resources": [
        {
            "apiVersion": "2018-05-01",
            "name": "[parameters('name')]",
            "location": "global",
            "type": "Microsoft.Maps/accounts",
            "properties": {},
            "sku": {
                "name": "[parameters('pricingSpec')]"
            }
        }
    ]
}
```

PR #468, opened 2018-05-30
api Management 2018-01-01 schema and tests

solankisamir:
2018-01-01 schema and tests


solankisamir:
@Tiano2017 can you take a look?

PR #469, opened 2018-06-02
Support user defined functions in root template.

Tiano2017:


simongdavies:
@Tiano2017  should parametervaluetype allow for some expressions?

Tiano2017:
@simongdavies one of the supported types is "string". that covers expressions because expressions are always string.

PR #474, opened 2018-06-11
Add schemas for AKS GA (managed clusters Kubernetes service)

### LINKED ISSUE #470: Missing schema for Microsoft.ContainerService/managedClusters for API Version 2018-03-31
Missing schema for Microsoft.ContainerService/managedClusters for API Version 2018-03-31

mboersma:
Generated with `autorest` from current master of azure-rest-api-specs.

Closes #470

cc: @ravbhatnagar @amanohar

### DIFF
--- a/schemas/2018-03-31/Microsoft.ContainerService.json
+++ b/schemas/2018-03-31/Microsoft.ContainerService.json
@@ -1,0 +1,720 @@
+{
+  "id": "https://schema.management.azure.com/schemas/2018-03-31/Microsoft.ContainerService.json#",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Microsoft.ContainerService",
+  "description": "Microsoft ContainerService Resource Types",
+  "resourceDefinitions": {
+    "managedClusters": {
+      "type": "object",
+      "properties": {
+        "name": {
+          "type": "string"
+        },
+        "type": {
+          "type": "string",
+          "enum": [
+            "Microsoft.ContainerService/managedClusters"
+          ]
+        },
+        "apiVersion": {
+          "type": "string",
+          "enum": [
+            "2018-03-31"
+          ]
+        },
+        "location": {
+          "type": "string",
+          "description": "Resource location"
+        },
+        "tags": {
+          "oneOf": [
+            {
+              "type": "object",
+              "additionalProperties": {
+                "type": "string"
+              }
+            },
+            {
+              "$ref": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression"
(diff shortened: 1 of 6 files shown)


schemas/2018-03-31/Microsoft.ContainerService.json
```diff
@@ -0,0 +1,720 @@
...
+      "required": [
+        "name",
+        "vmSize"
+      ],
+      "description": "Profile for the container service agent pool."
+    },
+    "ManagedClusterProperties": {
+      "type": "object",
+      "properties": {
+        "kubernetesVersion": {
```
amanohar: Should `fqdn` be here as well?

mboersma: Yes it should, and it's in the public swagger spec I used to generate this from. Weird. I wonder if it's confused by the definition being split between two files... I'll take a look.

mboersma: It looks like it's not included in the ARM resource schema because it's defined as `readOnly`. Same thing happens for `provisioningState` here. I assume this is correct--@ravbhatnagar?

ravbhatnagar:
@Tiano2017 - FYI for review and merge

Tiano2017:
@mboersma please add some simple tests in /tests folder.

mboersma:
@Tiano2017 the overall "ACS" API is split into three areas currently. From the public swagger:

- Microsoft.ContainerService/stable/2017-07-01/containerService.json
- Microsoft.ContainerService/stable/2018-03-31/managedClusters.json
- Microsoft.ContainerService/stable/2017-09-30/location.json

To update the deploymentTemplate.json files to point to a newer ARM schema, I would need to generate schemas for all of those and reference them. Is it possible to generate a single schema file from multiple inputs, or how should I handle this?

Tiano2017:
@mboersma I don't think AutoRest supports that today. You may need to do that manually if you want to merge. Actually it seems to me that you can just have different schema files since they have different API versions.

mboersma:
> you can just have different schema files

Thanks--yes, that's probably simplest. Let me rework this PR and add the ACS schema as well and update the deploymentTemplate.json.

ravbhatnagar:
@mboersma - any updates on this? We wanted to get a concrete status for fundamentals. Thanks!

mboersma:
@ravbhatnagar I'm working on it now and will have it done this week. Sorry for dropping the ball!

Tiano2017:
@mboersma please also add tests in the /tests folder.

mboersma:
@Tiano2017 I added some tests. Please let me know what else needs to be done.

PR #492, opened 2018-08-09
Update server properties

YingZhang-MSPBI:
Update AS schema and test cases according to RpPropertyBag.cs.

### DIFF
--- a/tools/package-lock.json
+++ b/tools/package-lock.json
@@ -1,0 +1,863 @@
+{
+  "name": "template-schema-validation-tools",
+  "version": "1.0.0",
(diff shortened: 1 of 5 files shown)


tools/package-lock.json
```diff
@@ -0,0 +1,863 @@
+{
+  "name": "template-schema-validation-tools",
```
TianoMS: >name [](start = 3, length = 4)

can we revert this file?

Tiano2017:
PR looks good. just one minor issue.

minghan:
Please hold on merging this. The api version is incorrect.

PR #505, opened 2018-09-06
policyDefinition rule schema (2018-05-01)

pilor:
Updating the policyDefinition rule schema to account for recent changes.  The latest api-version is 2018-05-01.  

This schema file is not a resource type in ARM.  It is purely the schema used for Azure Resource Policy rules and is consumed in the JSON editor in the Portal as well as by standalone tooling for policy authors.  

Compare  the second commit against the first to see the delta from the previous api-version (2016-12-01).

### DIFF
--- a/schemas/2018-05-01/policyDefinition.json
+++ b/schemas/2018-05-01/policyDefinition.json
@@ -1,0 +1,269 @@
+{
+  "id": "https://schema.management.azure.com/schemas/2016-12-01/policyDefinition.json#",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Policy Definition",
+  "description": "This schema defines Azure resource policy definition, please see https://azure.microsoft.com/en-us/documentation/articles/resource-manager-policy/ for more details.",
+  "type": "object",
+  "properties": {
+    "if": {
+      "oneOf": [
+        { "$ref": "#/definitions/condition" },
+        { "$ref": "#/definitions/operatorNot" },
+        { "$ref": "#/definitions/operatorAnyOf" },
+        { "$ref": "#/definitions/operatorAllOf" }
+      ]
+    },
+    "then": {
+      "type": "object",
+      "properties": {
+        "effect": {
+          "type": "string",
+          "enum": [ "append", "audit", "auditIfNotExists", "deny" ]
+        },
+        "details": {
+          "oneOf": [
+            { "$ref": "#/definitions/ifNotExistsDetails" },
+            { "$ref": "#/definitions/appendDetails" }
+          ]
+        }
+      },
+      "required": [ "effect" ],
+      "additionalProperties": false
+    }
+  },
+  "required": [ "if", "then" ],
+  "additionalProperties": false,
+  "definitions": {
+    "appendDetails": {
+      "type": "array",
+      "items": {
+        "properties": {
+          "field": {
+            "type": "string"
+          },
+          "value": {
+          }
+        },
+        "required": [ "field", "value" ],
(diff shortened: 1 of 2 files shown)


Tiano2017:
@bmoore-msft can you take a look at this one?

pilor:
> @bmoore-msft can you take a look at this one?

@bmoore-msft Can you review this when you get a chance?

schemas/2018-05-01/policyDefinition.json
```diff
@@ -0,0 +1,279 @@
...
+  "id": "https://schema.management.azure.com/schemas/2018-05-01/policyDefinition.json#",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Policy Definition",
+  "description": "This schema defines Azure resource policy definition, please see https://azure.microsoft.com/en-us/documentation/articles/resource-manager-policy/ for more details.",
+  "type": "object",
+  "properties": {
+    "if": {
+      "oneOf": [
+        { "$ref": "#/definitions/condition" },
+        { "$ref": "#/definitions/operatorNot" },
```
bmoore-msft: not a policy expert, but this seems cicular/re-entrant... the oneOfs here are just repeating the definition of the oneOf except for the condition.  If that's indeed valid, not sure why we need 3 of them

schemas/2018-05-01/policyDefinition.json
```diff
@@ -0,0 +1,279 @@
...
+        },
+        "existenceCondition": {
+          "oneOf": [
+            { "$ref": "#/definitions/condition" },
+            { "$ref": "#/definitions/operatorNot" },
+            { "$ref": "#/definitions/operatorAnyOf" },
+            { "$ref": "#/definitions/operatorAllOf" }
+          ]
+        },
+        "deployment": {
```
bmoore-msft: can we define this?  I looked for docs but didn't find anything on it...

pilor: let me see if I deep linking to the template deployment schema works well here

schemas/2018-05-01/policyDefinition.json
pilor: What each case can contain is a bit different.  It is hard to express it in other ways without looking like a mess.  This is the schema we use in the Ibiza UX JSON editor and I have used it extensively in VS code.  It does parse and validate successfully.

bmoore-msft: Ok - if it's working fine, let's not mess with it ;)

pilor:
@bmoore-msft This should be good to go now.  Good idea referencing the actual deployment schema

schemas/2018-05-01/policyDefinition.json
```diff
@@ -0,0 +1,284 @@
...
+          "type": "object",
+          "properties": {
+            "properties": {
+              "$ref": "https://schema.management.azure.com/schemas/2018-05-01/Microsoft.Resources.json#/definitions/DeploymentProperties"
```
TianoMS: >"$ref": "https://schema.management.azure.com/schemas/2018-05-01/Microsoft.Resources.json#/definitions/DeploymentProperties" [](start = 14, length = 123)

is this required? we actually don't support cross file references except the ones to the root schema. In the case of this file, it won't break because it's a standalone schema file. but just want to confirm this is definitely needed.

pilor: @TianoMS Its not required but it does make it easier to use.  Brian suggested it above.  Will it cause problems?  If so, I can revert it.

PR #512, opened 2018-09-26
Adding 2018-06-01 schemas for Compute

waynekuo:


schemas/2018-06-01/Microsoft.Compute.Extensions.json
```diff
@@ -0,0 +1,2025 @@
+{
+  "id": "https://schema.management.azure.com/schemas/2018-06-01/Microsoft.Compute.Extensions.json#",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Microsoft.Compute.Extensions",
+  "description": "Microsoft Compute Extensions Resource Types",
+  "definitions": {
```
TianoMS: >definitions [](start = 3, length = 11)

can we still put an empty "resourceDefinitions" property here?

tests/2018-06-01/Microsoft.Compute.tests.json
```diff
@@ -0,0 +1,1407 @@
+{
+  "tests": [
```
TianoMS: >tests [](start = 3, length = 5)

we have another set of tests in /tools/templateFiles, which are basically some test templates. Can we add the test resources to TestTemplate.json and TestTemplate.preview.json? 

Should be just a copy/paste task.

Tiano2017:
Looks good overall. I'll need to run some service CITs to make sure nothing breaks.

tests/2018-06-01/Microsoft.Compute.tests.json
waynekuo: Sorry, can you explain this comment a bit more? Do you want me to copy everything in this test file to TestTemplate.json and TestTemplate.preview.json?

PR #514, opened 2018-10-03
Corrected schema for SendGrid.Email/accounts resource

devoto13:
Updates are based on the errors received, when attempting to create a resource and [this blogpost](https://anthonychu.ca/post/arm-template-storage-sendgrid/).

### DIFF
--- a/schemas/2015-01-01/Sendgrid.Email.json
+++ b/schemas/2015-01-01/Sendgrid.Email.json
@@ -40,12 +40,12 @@
                 },
                 {
                   "enum": [
-                    "Free",
-                    "Bronze",
-                    "Silver",
-                    "Gold",
-                    "Platinum",
-                    "Premier"
+                    "free",
+                    "bronze",
+                    "silver",
+                    "gold",
+                    "platinum",
+                    "premier"
                   ]
                 }
               ],
--- a/schemas/2015-01-01/Sendgrid.Email.json
+++ b/schemas/2015-01-01/Sendgrid.Email.json
@@ -52,7 +52,16 @@
               "description": "Plan name"
             },
             "publisher": {
-              "type": "string",
+              "oneOf": [
+                {
+                  "$ref": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression"
+                },
+                {
+                  "enum": [
+                    "Sendgrid"
+                  ]
+                }
+              ],
               "description": "Publisher name"
             },
             "product": {
@@ -62,12 +71,7 @@
                 },
                 {
                   "enum": [
-                    "free",
-                    "bronze",
-                    "silver",
-                    "gold",
-                    "platinum",
-                    "2500000"
+                    "sendgrid_azure"
                   ]
                 }
               ],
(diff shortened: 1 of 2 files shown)


schemas/2015-01-01/Sendgrid.Email.json
```diff
@@ -78,7 +82,7 @@
               "description": "Promotion code"
             }
           },
-          "required": ["name", "product"],
+          "required": ["name", "publisher", "product"],
```
TianoMS: >publisher [](start = 32, length = 9)

is publisher required?

schemas/2015-01-01/Sendgrid.Email.json
```diff
@@ -40,19 +40,28 @@
...
             },
             "publisher": {
-              "type": "string",
+              "oneOf": [
+                {
+                  "$ref": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression"
+                },
+                {
+                  "enum": [
+                    "Sendgrid"
```
TianoMS: >Sendgrid [](start = 21, length = 8)

why is this one in upper case while everything else changes to lower case?

Tiano2017:
@devoto13 left a couple comments. can  you take a look?

schemas/2015-01-01/Sendgrid.Email.json
devoto13: This is how Azure Portal sends it to the API. I've tested the lowercase `sendgrid` and it works as well, but I would prefer to keep it same as it is done in the Azure Portal.

As for changing plan names to lowercase - there are no plans named in title case. Took me a LOT of time to figure out that the problem was because of the not existing plan name and not because of the wrong market:

```
$ cat test.json
{
    "$schema": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
    "contentVersion": "1.0.0.0",
    "parameters": {},
    "variables": {},
    "resources": [
        {
            "name": "plus1CloudTest",
            "type": "Sendgrid.Email/accounts",
            "location": "[resourceGroup().location]",
            "apiVersion": "2015-01-01",
            "plan": {
                "name": "Free",
                "publisher": "Sendgrid",
                "product": "sendgrid_azure"
            },
            "properties": {
                "password": "123ASDfgh",
                "acceptMarketingEmails": false,
                "email": ""
            }
        }
    ]
}
$ az --verbose group deployment create --resource-group plus1tools-dev-rg --mode Incremental --template-file test.json
Accepted: test (Microsoft.Resources/deployments)
Failed: plus1CloudTest (Sendgrid.Email/accounts)
Deployment failed. Correlation ID: 7826a232-8297-4045-9aee-28d787d9882d. {
  "error": {
    "code": "ResourcePurchaseValidationFailed",
    "message": "User failed validation to purchase resources. Error message: '{\"ErrorDescription\":\"[\\\"CustomerDeniedServicePlanAccess The Offer is not sold in the account market (DK).\\\"]\",\"CultureInvariantErrorCode\":\"BadRequest\",\"ActivityId\":\"491ba1ee-1b7f-414c-903f-e3d4e16734ea\"}'"
  }
}
```

Same for the `product` field. Previous value were wrong as they don't exist. The new value is what allows to create the resource successfully and is what Azure Portal sends to the API.

Tiano2017:
PR looks good now. will merge.

PR #515, opened 2018-10-03
Add schema for Microsoft.Media for 2018-07-01

quintinb:
Adding the ARM Template schema for Microsoft.Media for 2018-07-01.  I updated the two deploymentTemplate.json files and ran the two validation scripts.  I added at least one negative test case and one positive test case for each new resource type.  I tried to add multiple positive examples for the more complex cases.

Feedback welcome.


TianoMS:
>           "format": "duration",

the "duration" format is not supported by Visual Studio. can we remove them?

---
Refers to: schemas/2018-07-01/Microsoft.Media.json:1788 in 8f79cd1. [](commit_id = 8f79cd16ec5d2aa08eb4c10ed574cce7d61b4a89, deletion_comment = False)

schemas/2014-04-01-preview/deploymentTemplate.json
```diff
@@ -275,6 +275,18 @@
                   { "$ref": "https://schema.management.azure.com/schemas/2018-06-30/Microsoft.Automation.json#/resourceDefinitions/automationAccounts_python2Packages" },
                   { "$ref": "https://schema.management.azure.com/schemas/2018-06-30/Microsoft.Automation.json#/resourceDefinitions/automationAccounts_runbooks" },
                   { "$ref": "https://schema.management.azure.com/schemas/2015-10-01/Microsoft.Media.json#/resourceDefinitions/mediaServices" },
+                  { "$ref": "https://schema.management.azure.com/schemas/2018-07-01/Microsoft.Media.json#/resourceDefinitions/mediaServices" },
```
TianoMS: >mediaServices [](start = 126, length = 13)

please use consistent case here and in the schema file.

quintinb: Fixed.  I removed the duration format mentioned above and standardized the casing of the definitions in the deployment templates, schema, and the associated tests.

TianoMS: >mediaServices [](start = 126, length = 13)

by "case" I mean, here you have capital S, but in the schema file you have "mediaservices" where everything is lower case.

TianoMS:
>     "mediaservices": {

here, the resource type name is all lower case. 

---
Refers to: schemas/2018-07-01/Microsoft.Media.json:7 in 4632a38. [](commit_id = 4632a38c362fb9ea9405a7cdf59f9dd4470e3350, deletion_comment = False)

schemas/2014-04-01-preview/deploymentTemplate.json
quintinb: I checked the existing 2015-10-01 schema and the "mediaservices" (all lower case) version was a mistake.  I have fixed everything to be mixed case ("mediaServices") in the schema file.

PR #520, opened 2018-10-17
Add Microsoft.AAD Domain Services Schemas

jihochang:


schemas/2017-06-01/Microsoft.AAD.json
```diff
@@ -0,0 +1,239 @@
...
+            "Microsoft.AAD/domainServices"
+          ]
+        },
+        "apiVersion": {
+          "type": "string",
+          "enum": [
+            "2017-06-01"
+          ]
+        },
+        "domainName": {
```
TianoMS: >domainName [](start = 9, length = 10)

Is this correct? Isn't this supposed to be inside the "properties" bag? Only "name", "type", "apiVersion" and a few others are allowed at top level.

jihochang: I have fixed this.

PR #527, opened 2018-11-01
Regenerated schema based on Swagger and added test.

huizlAzure:


schemas/2018-07-01/Microsoft.Storage.json
```diff
@@ -105,6 +115,64 @@
       ],
       "description": "Microsoft.Storage/storageAccounts"
     },
+    "storageAccounts_blobServices": {
```
TianoMS: >storageAccounts_blobServices [](start = 5, length = 28)

do you need to add this to 2015-01-01/deploymentTemplate.json and 2014-04-01-preview/deploymentTemplate?

huizlAzure: Blob services APIs are supported only in 2018-07-01 and later SRP API versions, so I think we do not want to add it to those versions here?

I actually am not sure why all the existing objects have 2015-01-01 in 'oneOf', on the side of the type ref. I'm new to Azure schema. Any comments/suggestions?

huizlAzure: Tiano: I've added BlobServices in those deployment template files in latest commit. Kindly take a look.

tests/2018-07-01/Microsoft.Storage.tests.json
```diff
@@ -197,6 +197,67 @@
           "immutabilityPeriodSinceCreationInDays": 3
         }
       }
+    },
+    {
+      "name": "storageAccountBlobServices",
```
TianoMS: >storageAccountBlobServices [](start = 15, length = 26)

test case is failing.

tests/2018-07-01/Microsoft.Storage.tests.json
```diff
@@ -197,6 +197,67 @@
...
+                "allowedMethods": [
+                  "GET",
+                  "PUT"
+                ],
+                "maxAgeInSeconds": 500,
+                "exposedHeaders": [
+                  "x-ms-meta-customheader",
+                  "x-ms-meta-data*"
+                ],
+                "allowedHeaders: [
```
TianoMS: >: [](start = 31, length = 1)

a double quote is missing here.

tests/2018-07-01/Microsoft.Storage.tests.json
huizlAzure: Fixed missing double quote. Where do I find test results in the PR?

PR #533, opened 2018-11-08
Removed VA policy schema

t-robalt:
Removed DatabaseVulnerabilityAssessmentProperties and all elements dependent upon it.

### DIFF
--- a/schemas/2017-03-01-preview/Microsoft.Sql.json
+++ b/schemas/2017-03-01-preview/Microsoft.Sql.json
@@ -175,9 +175,6 @@
                 "$ref": "#/definitions/servers_databases_backupLongTermRetentionPolicies_childResource"
               },
               {
-                "$ref": "#/definitions/servers_databases_vulnerabilityAssessments_childResource"
-              },
-              {
                 "$ref": "#/definitions/servers_databases_auditingSettings_childResource"
               },
               {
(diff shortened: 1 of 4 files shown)


schemas/2017-03-01-preview/Microsoft.Sql.json
```diff
@@ -326,36 +326,6 @@
       ],
       "description": "Microsoft.Sql/servers/databases/extendedAuditingSettings"
     },
-    "servers_databases_vulnerabilityAssessments": {
```
TianoMS: >servers_databases_vulnerabilityAssessments [](start = 5, length = 42)

this is a breaking change. Is the API already removed from service?

TianoMS: >servers_databases_vulnerabilityAssessments [](start = 5, length = 42)

this needs to be removed from 2015-01-01/deploymentTemplate and 2014-04-01-preview/deploymentTemplate

t-robalt: No, and it won't.
This is a temporary fix for a problem we are working on.

Tiano2017:
@jaredmoo  I assume you guys are going to track the issue and add the changes back later.

PR #540, opened 2018-11-22
Add Microsoft.NetApp schema

draupnirgudmundsson:
Adding the schema to work with Azure CloudVolumes (Microsoft.NetApp)


ravbhatnagar:
Looks good.

PR #545, opened 2018-12-06
Schema for migration APIs

lshai06:
This PR captures the schema for migration APIs that will be exposed by Azure Site Recovery (ASR). ASR already exposes APIs for replication/failover, and we have tried to adhere closely to the same while adding support for migration as a scenario.
The overall model hierarchy of (vault -> fabric -> container) has been retained, and migration item will now fall below container. New APIs added for migration items are in line with the existing replication protected items APIs. Migration items are created by an Enable operation. Once a migration item is created, we can invoke a test migrate and cleanup on it, or a migrate and complete migration on it, to confirm that the migration is completed. The migration item can be removed/force removed by a Disable/Delete operation.
We are also exposing the put for ReplicationRecoveryServicesProviders, which we had missed earlier along with its get and delete


lshai06:
This is just regenerating the schemas based on the latest swagger.
[Codeflow ](http://codeflow/extensions/launcher.html?server=https:%2f%2fmicrosoft.visualstudio.com%2fdefaultcollection&projectId=4e9ae0d6-7a8d-4dd3-91c8-887ebe98162e&reviewId=2705685&projectshortname=HSDP)may help to see changes more clearly.

lshai06:
@ravbhatnagar Can you please take a look? This is the schema for migration swagger.

schemas/2018-01-10/Microsoft.RecoveryServices.json
```diff
@@ -1981,58 +2232,147 @@
...
-        },
-        "apiVersion": {
-          "type": "string",
-          "enum": [
-            "2018-01-10"
+      "oneOf": [
+        {
+          "allOf": [
+            {
+              "$ref": "#/definitions/A2AContainerMappingInput"
```
TianoMS: >A2AContainerMappingInput [](start = 37, length = 24)

this is not defined..

Tiano2017:
@lshai06 one comment please take a look. Also please sync the latest master.

PR #547, opened 2018-12-12
Add data connections type to the schema

radennis:
Update the schema with data connections type


radennis:
@Tiano2017 could you please review? 
Also a question, is it possible to generate the schema based on the swagger or do I have to manually write it?

radennis:
@ravbhatnagar Can you take a look?

radennis:
@ravbhatnagar could you please rake a look?

PR #552, opened 2018-12-20
Minor updates to the 2018-07-01 API

BrianBlum:
The properties on the presentationTimeRange are no longer required.

### DIFF
--- a/schemas/2018-07-01/Microsoft.Media.json
+++ b/schemas/2018-07-01/Microsoft.Media.json
@@ -1785,14 +1785,17 @@
         },
         "relativeBeginDate": {
           "type": "string",
+          "format": "duration",
           "description": "The relative begin date of license."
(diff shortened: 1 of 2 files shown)


schemas/2018-07-01/Microsoft.Media.json
```diff
@@ -1864,6 +1867,7 @@
       "properties": {
         "firstPlayExpiration": {
           "type": "string",
+          "format": "duration",
```
TianoMS: >"format": "duration", [](start = 10, length = 21)

duration format is not supported by Visual Studio... It's an issue of VS but right now we have to work around for that...

antmarti-microsoft: @TianoMS Worth adding a validation to catch this?

PR #559, opened 2019-01-08
Added Microsoft.Peering schema and tests

rileymckenna:
passed 1/1.
@Tiano2017


tools/package-lock.json
```diff
@@ -0,0 +1,754 @@
+{
+  "name": "template-schema-validation-tools",
```
TianoMS: >template-schema-validation-tools [](start = 11, length = 32)

do we need this new file?

PR #571, opened 2019-02-04
Remove availability zones

radennis:


radennis:
@Tiano2017 could you please take a look?

PR #577, opened 2019-02-28
Removed required field from VulnerabilityAssessmentProperties

t-robalt:
Removed the requirement of StorageContainerPath in the properties of Server level and manged Instance level vulnerability assessment.


t-robalt:
@Tiano2017 Can you please review this, or address someone else?
This fixes a small bug in prod, so please review it when possible.

PR #583, opened 2019-03-21
Updated Network 2018-12-01 from the latest specs

EvgenyAgafonchikov:


schemas/2018-12-01/Microsoft.Network.json
```diff
@@ -89,6 +89,67 @@
...
+    "ApplicationGatewayWebApplicationFirewallPolicies": {
+      "type": "object",
+      "properties": {
+        "name": {
+          "type": "string"
+        },
+        "type": {
+          "type": "string",
+          "enum": [
+            "Microsoft.Network/ApplicationGatewayWebApplicationFirewallPolicies"
```
TianoMS: >ApplicationGatewayWebApplicationFirewallPolicies [](start = 31, length = 48)

do we need to add it to the root schemas?

TianoMS: >ApplicationGatewayWebApplicationFirewallPolicies [](start = 31, length = 48)

add test?

EvgenyAgafonchikov: updated

EvgenyAgafonchikov: Added some tests.
It would be great if related specs developer could review these tests and suggest improvements or add more there if required. @azhao155, could you please take a look?

PR #585, opened 2019-03-28
Add autogenerated schemas

antmarti-microsoft:
Autogenerated schemas taken from https://github.com/fearthecowboy/generated-schemas/tree/a03b16aa47af7253289aa8cd5d93f969616474db

The schemas that have been selected are only for RPs with 0 existing schemas in this repo.


schemas/2015-01-01/deploymentTemplate.json
```diff
@@ -916,7 +916,73 @@
                   { "$ref": "https://schema.management.azure.com/schemas/2018-06-01/Microsoft.Security.json#resourceDefinitions/autoProvisioningSettings"},
                   { "$ref": "https://schema.management.azure.com/schemas/2018-06-01/Microsoft.Security.json#resourceDefinitions/advancedThreatProtectionSettings"},
                   { "$ref": "https://schema.management.azure.com/schemas/2018-06-01/Microsoft.Security.json#resourceDefinitions/informationProtectionPolicies"},
-                  { "$ref": "https://schema.management.azure.com/schemas/2019-03-01/Microsoft.Insights.json#/resourceDefinitions/actionGroups" }
+                  { "$ref": "https://schema.management.azure.com/schemas/2019-03-01/Microsoft.Insights.json#/resourceDefinitions/actionGroups" },
```
TianoMS: >{ "$ref": "https://schema.management.azure.com/schemas/2019-03-01/Microsoft.Insights.json#/resourceDefinitions/actionGroups" }, [](start = 18, length = 127)

I think we don't want to add to the existing root schemas. The new schemas are only for export template, and we are to create a new root schema for that.

antmarti-microsoft: Updated to create new root schema. Please take another look.

Tiano2017:
@antmarti-microsoft can we add one test template in tools/TemplateFiles that points to the new root schema?

PR #587, opened 2019-03-30
Fixing JSON schema conformance

wpbrown:
Fixing JSON schema spec conformance problems. These problems cause exceptions when trying to load the ARM schema in Newtonsoft Json.NET Schema.


schemas/2015-01-01/deploymentTemplate.json
```diff
@@ -904,25 +904,25 @@
...
-                  { "$ref": "https://schema.management.azure.com/schemas/2018-06-01/Microsoft.Security.json#resourceDefinitions/pricings"},
-                  { "$ref": "https://schema.management.azure.com/schemas/2018-06-01/Microsoft.Security.json#resourceDefinitions/securityContacts"},
-                  { "$ref": "https://schema.management.azure.com/schemas/2018-06-01/Microsoft.Security.json#resourceDefinitions/workspaceSettings"},
-                  { "$ref": "https://schema.management.azure.com/schemas/2018-06-01/Microsoft.Security.json#resourceDefinitions/autoProvisioningSettings"},
-                  { "$ref": "https://schema.management.azure.com/schemas/2018-06-01/Microsoft.Security.json#resourceDefinitions/advancedThreatProtectionSettings"},
-                  { "$ref": "https://schema.management.azure.com/schemas/2018-06-01/Microsoft.Security.json#resourceDefinitions/informationProtectionPolicies"},
+                  { "$ref": "https://schema.management.azure.com/schemas/2015-03-01-preview/Microsoft.HDInsight.json#/resourceDefinitions/clusters" },
```
TianoMS: >/ [](start = 117, length = 1)

I assume this is the focus of the change?

schemas/2015-01-01/deploymentTemplate.json
```diff
@@ -904,25 +904,25 @@
...
+                  { "$ref": "https://schema.management.azure.com/schemas/2017-08-01-preview/Microsoft.Security.json#/resourceDefinitions/informationProtectionPolicies"},
+                  { "$ref": "https://schema.management.azure.com/schemas/2018-06-01/Microsoft.Security.json#/resourceDefinitions/pricings"},
+                  { "$ref": "https://schema.management.azure.com/schemas/2018-06-01/Microsoft.Security.json#/resourceDefinitions/securityContacts"},
+                  { "$ref": "https://schema.management.azure.com/schemas/2018-06-01/Microsoft.Security.json#/resourceDefinitions/workspaceSettings"},
+                  { "$ref": "https://schema.management.azure.com/schemas/2018-06-01/Microsoft.Security.json#/resourceDefinitions/autoProvisioningSettings"},
+                  { "$ref": "https://schema.management.azure.com/schemas/2018-06-01/Microsoft.Security.json#/resourceDefinitions/advancedThreatProtectionSettings"},
+                  { "$ref": "https://schema.management.azure.com/schemas/2018-06-01/Microsoft.Security.json#/resourceDefinitions/informationProtectionPolicies"},
```
TianoMS: >resourceDefinitions [](start = 109, length = 19)

do we need to update 2014-04-01-preview/deploymentTemplate.json as well?

Tiano2017:
@wpbrown Thanks for making the change. there are some conflicts with origin/master. I also left two comment. Can you take a look?

schemas/2015-01-01/deploymentTemplate.json
wpbrown: Yes the only change in each line is the new `/` after the `#`. Without it, a reference to an "id" is implied.

schemas/2015-01-01/deploymentTemplate.json
wpbrown: I wasn't familiar with that one, but it would seem so. I'll update it with the same change.

wpbrown:
@Tiano2017 I merged master and updated 2014-04-01-preview as well.

PR #622, opened 2019-06-17
added cuidv2 handler

quinn-mchugh:
The new version of CreateUIDefinition implements a new potential handler. Updating the latest 0.1.2 schema to be aware of the new handler value.


quinn-mchugh:
The new version of CreateUIDefinition implements a new potential handler. Updating the latest 0.1.2 schema to be aware of the new handler value.

PR #623, opened 2019-06-18
Add resource type definitions for 2015-04-08

bizarreWizard:


schemas/2019-04-01/autogeneratedResources.json
```diff
@@ -10,6 +10,36 @@
     },
     {
       "oneOf": [
```
antmarti-microsoft: Please can you also:
1. Remove any references to this schema from `schemas/2019-04-01/deploymentTemplate.json` (as this is now covered by `schemas/2019-04-01/autogeneratedResources.json`).
2. Manually add the new types if required to `schemas/2014-04-01-preview/deploymentTemplate.json` and `schemas/2015-01-01/deploymentTemplate.json`.
3. Add CosmosDb to the list of autogenerated schema RPs (in `generator/scripts/constants.ps1` - can add the following line: `'cosmos-db/resource-manager/Microsoft.DocumentDB' = @('*')`.

PR #624, opened 2019-06-19
EventGrid: Add templates and related test cases for 2019-06-01 API version

ahamad-MS:
EventGrid: Add templates and related test cases for 2019-06-01 API version


schemas/2019-04-01/autogeneratedResources.json
```diff
@@ -231,6 +231,15 @@
         },
         {
           "$ref": "https://schema.management.azure.com/schemas/2019-04-01/Microsoft.DevSpaces.json#/resourceDefinitions/controllers"
+        },
```
antmarti-microsoft: `schemas/2019-04-01/deploymentTemplate.json` pulls in this file, so by including in both places this will cause a duplicate definition. Since some manual work was required to add "eventSubscriptions", please remove the changes to the "autogeneratedResources.json" file, and we can track this schema as manually created for now.

antmarti-microsoft:
Just FYI we have https://github.com/Azure/autorest.azureresourceschema/issues/27 open to track the schema generator failing to generate a schema for the eventSubscriptions resource.

ahamad-MS:
@antmarti-microsoft  Thanks Anthony for the info and opening the issue for eventsubscription generator.

PR #639, opened 2019-07-02
Adding new properties enableCassandraConnector and connectorOffer to microsoft.documentdb version 2015-04-08

bhba-coder:


schemas/2015-04-08/Microsoft.DocumentDB.json
```diff
@@ -869,6 +869,31 @@
             }
           ],
           "description": "Enables the account to write in multiple locations"
+        },
+        "enableCassandraConnector": {
```
antmarti-microsoft: I cannot find these two new properties in the [current CosmosDb swagger spec](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/cosmos-db/resource-manager/Microsoft.DocumentDB/stable/2015-04-08/cosmos-db.json). How was this schema generated? Will these properties be added to the swagger spec?

bhba-coder: https://github.com/Azure/azure-rest-api-specs/pull/6550

PR #641, opened 2019-07-02
README update

putical:
Added some information about having the public SSH key configured


antmarti-microsoft:
@putical I think something has gone badly wrong with this PR! It looks like your fork of the repo is some way behind master, and also that you're trying to merge into the wrong branch (you need to select "master" to merge into).

If you're only making a small change, I'd recommend resetting your fork, and then re-doing your change on top of it.

putical:
Hi Anthony, sorry,  my bad. please ignore the PR I made, I meant to do that on the forked repo as a test. Again, my apologies!

PR #642, opened 2019-07-09
Deployment Template for 2019-03-01-hybrid for Azure Stack 

t-anba:


schemas/2019-03-01-hybrid/deploymentTemplate.json
```diff
@@ -0,0 +1,3639 @@
+﻿{
+  "id": "https://raw.githubusercontent.com/t-anba/azure-resource-manager-schemas/t-anba-AzsDeploymentTemplate-final/schemas/2019-04-01/deploymentTemplate.json#",
```
antmarti-microsoft: Make sure you replace this with the correct URL - I think in this case it should be `https://schema.management.azure.com/schemas/2019-03-01-hybrid/deploymentTemplate.json#`

schemas/2019-03-01-hybrid/deploymentParameters.json
```diff
@@ -0,0 +1,57 @@
+{
+  "id": "https://schema.management.azure.com/schemas/2019-04-01/deploymentParameters.json#",
```
antmarti-microsoft: Make sure URL is correct for this file

schemas/2019-03-01-hybrid/deploymentTemplate.json
```diff
@@ -0,0 +1,3639 @@
...
+    }
+  },
+  "additionalProperties": false,
+  "required": [
+    "$schema",
+    "contentVersion",
+    "resources"
+  ],
+  "definitions": {
+    "2019-03-01-hybrid-profile-resource": {
```
antmarti-microsoft: If you have a script to regenerate this file from the stack-supported list of resources, please could you check this script into this repo?

schemas/2019-03-01-hybrid/deploymentTemplate.json
```diff
@@ -0,0 +1,3639 @@
...
+      "oneOf": [
+        {
+          "$ref": "https://schema.management.azure.com/schemas/2016-10-01/Microsoft.KeyVault.json#/resourceDefinitions/vaults"
+        },
+        {
+          "$ref": "https://schema.management.azure.com/schemas/2016-10-01/Microsoft.KeyVault.json#/resourceDefinitions/vaults_secrets"
```
antmarti-microsoft: Please ensure we have a test that will break if one of these referenced resources is changed (for example if `https://schema.management.azure.com/schemas/2016-10-01/Microsoft.KeyVault.json#/resourceDefinitions/vaults_secrets` is deleted or renamed).

Please also ensure we have a clear list of instructions/script that can be run to unblock RPs if this test fails.

schemas/2019-03-01-hybrid/deploymentTemplate.json
t-anba: I dont currently have a script... should I make one ?

antmarti-microsoft: It would certainly be helpful for maintainability for both ARM & Azure Stack going forwards! How did you generate the list of resources? The scripts in ./generator/scripts in this repo are a good starting point and I can give you some other pointers.

t-anba: I did it by hand for adding the resources one by one. To write the script for stack to generate the deployment Template i would need to first grab the api profile comb to see what is supported and then put in the ref to the supported apiprofiles

PR #644, opened 2019-07-11
[Compute] Updating 2019-03-01 schemas for Compute(dedicated host api)

mabhard:


hyonholee:
LGTM

hyonholee:
@antmarti-microsoft   Please review this PR.

antmarti-microsoft:
Should these new resource types be added to the root schemas:
schemas\2014-04-01-preview\deploymentTemplate.json
schemas\2015-01-01\deploymentTemplate.json
schemas\2019-04-01\deploymentTemplate.json

Please could you add/update tests for new/changed resource types.

mabhard:
The files already contain the resources types for version "2019-03-01".

schemas/2019-03-01/Microsoft.Compute.json
```diff
@@ -259,6 +384,153 @@
...
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            },
+            {
+              "$ref": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression"
+            }
+          ],
+          "description": "Availability Zone to use for this host group � only single zone is supported. The zone can be assigned only during creation. If not provided, the group supports all zones in the region. If provided, enforce each host in the group is in the same zone."
```
huangpf: There seems to be a non-ASCII character in this line? Is it OK?

group � only

hyonholee: @mabhard  Can you remove the character in this PR?  It looks like we need to update the Swagger by removing it too.

mabhard: Removed.
Should I remove in swagger too?

PR #652, opened 2019-07-25
add 2019-06-01 schemas for container service

xizhamsft:
adding 2019-06-01 schemas for container service. The goal is to fix the issue of incorrect template exported from azure portal.

### DIFF
--- a/schemas/2019-04-01/autogeneratedResources.json
+++ b/schemas/2019-04-01/autogeneratedResources.json
@@ -261,6 +261,9 @@
         },
         {
           "$ref": "https://schema.management.azure.com/schemas/2019-04-01/Microsoft.DevSpaces.json#/resourceDefinitions/controllers"
+        },
+        {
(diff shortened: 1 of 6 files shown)


schemas/2019-04-01/autogeneratedResources.json
```diff
@@ -261,6 +261,12 @@
         },
         {
           "$ref": "https://schema.management.azure.com/schemas/2019-04-01/Microsoft.DevSpaces.json#/resourceDefinitions/controllers"
+        },
+        {
+          "$ref": "https://schema.management.azure.com/schemas/2019-06-01/Microsoft.ContainerService.json#/resourceDefinitions/managedClusters"
+        },
+        {
+          "$ref": "https://schema.management.azure.com/schemas/2019-06-01/Microsoft.ContainerService.json#/resourceDefinitions/managedClusters_agentPools"
```
antmarti-microsoft: Does managedClusters_agentPools also need to be added to the other root schemas?

xizhamsft: Probably not. I will remove this. It is an untracked child resource.

PR #656, opened 2019-07-30
[Do not merge] Add schema for Network 2015-05-01-preview

number213:


number213:
Requires splitting multi-API schemas intro separate ones, will do in another PR

PR #668, opened 2019-08-05
Create swagger definitions for new Microsoft.OperationalInsights reso…

olterner:
…urce - Cluster


anthony-c-martin:
I think you've got the wrong repo - swagger should be checked in to https://github.com/Azure/azure-rest-api-specs.

If this has already been checked into the swagger repo, then see [here](https://github.com/Azure/azure-resource-manager-schemas#autogenerating-from-swagger) for info on generating the schema.

PR #671, opened 2019-08-07
Updating to 2019-06-01-preview for Peering

rileymckenna:
Updated Data model for Peering and PeerAsn and tests to accompany.


rileymckenna:
i had a type in the schema `peerAsns` now resolved.

PR #673, opened 2019-08-08
Add common definition files

anthony-c-martin:


schemas/common/definitions.json
```diff
@@ -0,0 +1,257 @@
...
+      "type": "object",
+      "properties": {
+        "name": {
+          "type": "string",
+          "description": "Name of the copy"
+        },
+        "count": {
+          "oneOf": [
+            {
+              "$ref": "https://schema.management.azure.com/schemas/common/definitions.json#/definitions/expression"
```
Tiano2017: >https://schema.management.azure.com/schemas/common/definitions.json#/definitions/expression [](start = 23, length = 91)

use relative path?

schemas/common/definitions.json
```diff
@@ -0,0 +1,257 @@
+{
+  "id": "https://schema.management.azure.com/schemas/common/definitions.json#",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "definitions",
+  "description": "Common definitions",
+  "definitions": {
```
Tiano2017: >definitions [](start = 3, length = 11)

we have something in deploymentTemplateParameters.json. Do we need to copy that over?

anthony-c-martin: No - that was a deliberate omission. All of the definitions I could find in deploymentTemplateParameters.json were specific to the api-version.

schemas/common/definitions.json
anthony-c-martin: Yup, let me fix that.

schemas/common/autogeneratedResources.json
```diff
@@ -0,0 +1,16 @@
+{
+  "id": "https://schema.management.azure.com/schemas/common/autogeneratedResources.json#",
```
majastrz: >id [](start = 3, length = 2)

Do we actually want to call the new schema "autogenerated"? From the customer POV, won't that sound a little weird?

schemas/common/manuallyAddedResources.json
```diff
@@ -0,0 +1,15 @@
+{
+  "id": "https://schema.management.azure.com/schemas/common/manuallyAddedResources.json#",
```
majastrz: >id [](start = 3, length = 2)

Same comment about naming here as well.

PR #686, opened 2019-08-14
Add missing definitions for Microsoft.Network 2018-04-01

number213:


number213:
@ifeoluwaokunoren could you please take a look?

Issue #257, opened 2016-11-03
Microsoft.Sql schema 2014-04-01-preview missing Collation


OzBob:
https://github.com/Azure/azure-resource-manager-schemas/blob/15534a69026cd9357b3db825a6cdb51457406233/schemas/2014-04-01-preview/Microsoft.Sql.json

does not contain actual Collations supported in Azure.

SQL_Latin1_General_CP1_CI_AS

Is not listed but is valid during ARM deployment.

yet, when corrected to be valid (according to the schema results in ARM in VS) the deployment fails, with the error: "'SQL_Latin1_General_Cp1_CI_AS_KI_WI' is not a supported collation."

OzBob:
The big question is how do the team that 'owns' SQL Server on Azure who set the concrete supported Collations know to update these schema files?

ifeoluwaokunoren:
We're deprecating the 2014-04-01-preview schema. Customers are encouraged to use  **2014-04-01** api-version or later preview version instead.

PR #696, opened 2019-08-20
Remove 2014-04-01-preview Microsoft.Sql schemas

jaredmoo:
This version was replaced long ago by 2014-04-01 (stable). Documenting this version is not sustainable because it predates swagger spec and we have no plans to add swagger spec for it. We encourage customers to switch to 2014-04-01 (stable) or later preview version.

This removal means that #257 will not be fixed.


anthony-c-martin:
@jaredmoo Looks like you also need to remove references to this resource from the following test files:

/tools/templateTests/Template_1.2014-01-01-preview.json
/tools/templateTests/Template_1.2015-01-01.json
/tools/templateTests/Template_1.2019-04-01.json
/tools/templateTests/Template_2.2014-01-01-preview.json

PR #699, opened 2019-08-21
Re-generate schema for Network 2019-04-01/2019-06-01

number213:


number213:
I've added preprocessing to our script so properties like "etag" and "provisioningState" would be marked as read-only (sometimes in Swagger they are not) and re-generated schema

number213:
@anthony-c-martin @ifeoluwaokunoren please take a look

PR #701, opened 2019-08-27
Adding GA version to  HealthcareApis

poadhika:
There is a schema already for HealthcareApis preview version, which was 2018-08-20-preview.  This change is for upgrading to GA version which is 2019-09-16.


schemas/2019-09-16/Microsoft.HealthcareApis.json
```diff
@@ -0,0 +1,336 @@
...
+        "type": {
+          "type": "string",
+          "enum": [
+            "Microsoft.HealthcareApis/services"
+          ]
+        },
+        "apiVersion": {
+          "type": "string",
+          "enum": [
+            "2018-08-20-preview",
```
anthony-c-martin: I think this is the reason for your test failure - this API version is defined in multiple places, so the schema parser cannot uniquely identify the resource.

This should just reference a single API version - i.e:

```
        "apiVersion": {
          "type": "string",
          "enum": [
            "2019-09-16"
          ]
        },
```

Issue #702, opened 2019-08-27
 apiProfile does not support latest values 


StephenWeatherford:
Reported by customer here: https://github.com/microsoft/vscode-azurearmtools/issues/279
> ### Problem
> The linter does not support the latest allowed values for the **apiProfiles** property.
> 
> ![image](https://user-images.githubusercontent.com/7881529/63770082-a5b16880-c8d4-11e9-80f7-aadba96a88cf.png)
> 
> ### Expected Behavior
> The linter should allow all possible values that are available in the Git repository for the Azure REST API (https://github.com/Azure/azure-rest-api-specs/tree/master/profile).

jcoeltjen:
> > This profile 2019-07-01-profile is not supported from ARM's perspective. It will be used for SDKs/PowerShell.
> > @StephenWeatherford, please close this issue.
> 
> This should be documented. What is the latest version allowed in an ARM template?

You can usually find the available template versions in the ARM template reference: https://docs.microsoft.com/en-us/azure/templates/

jihu:
> You can usually find the available template versions in the ARM template reference: 
> https://docs.microsoft.com/en-us/azure/templates/

Well, I can't really find any information about apiProfile versions there. Or am I missing something? If you can find it, maybe you can help me out with a direct link to that page?

jcoeltjen:
> > You can usually find the available template versions in the ARM template reference:
> > https://docs.microsoft.com/en-us/azure/templates/
> 
> Well, I can't really find any information about apiProfile versions there. Or am I missing something? If you can find it, maybe you can help me out with a direct link to that page?

Of course you are correct. I confused apiProfile and apiVersions.
The ARM documentation also only points to this git repository (see link in first post).
In this case I follow your idea: It should be documented which each api profile is intended for.

PR #703, opened 2019-08-28
Added missing apiProfile enum values

LasseSviland:
Added missing versions for apiProfile. 
 Issue: #702


anthony-c-martin:
We've had confirmation from the Azure Stack team that these apiProfile versions are not designed to be used with ARM templates, so I'm closing out this PR.

PR #705, opened 2019-09-09
Use latest Autorest V3 binaries for autogen

anthony-c-martin:


schemas/2015-08-01-preview/Microsoft.Portal.json
```diff
@@ -196,7 +206,11 @@
           "oneOf": [
             {
               "type": "object",
-              "additionalProperties": {}
+              "additionalProperties": {
+                "type": "object",
+                "properties": {}
+              },
+              "properties": {}
```
Tiano2017: is it expected that we want both "properties" and "additionalProperties"?

anthony-c-martin: Yeah, both are valid. It's just a little more explicit.

PR #713, opened 2019-09-18
Add support for differently-scoped resources

anthony-c-martin:
1. Add support for differently-scoped resources to autogeneration.
2. Add means for overriding unknown-scoped resources to whitelist.
3. Fail generate-all if expected whitelist files not found, namespace mismatch, or unresolvable scope.


anthony-c-martin:
Prerequisite for onboarding EventGrid. since they have a resource type which is valid at multiple scopes:
https://github.com/Azure/azure-rest-api-specs/blob/02da14696141d03069c47305a47a41a2d5acc0c3/specification/eventgrid/resource-manager/Microsoft.EventGrid/stable/2019-06-01/EventGrid.json#L698

generator/models.ts
```diff
@@ -0,0 +1,26 @@
+enum ScopeType {
+    None = 0,
+    Unknown = 1 << 0,
```
Tiano2017: >Unknown [](start = 4, length = 7)

Is there an example of "Unknown" type?

generator/models.ts
```diff
@@ -0,0 +1,26 @@
...
+    Tenant = 1 << 1,
+    Subcription = 1 << 2,
+    ResourceGroup = 1 << 3,
+    ManagementGroup = 1 << 4,
+    Extension = 1 << 5,
```
Tiano2017: >Extension [](start = 4, length = 9)

and "Extension" needs to be have its own type?

generator/models.ts
anthony-c-martin: https://github.com/Azure/azure-rest-api-specs/blob/f6e96392dd7f1069b6d600cfda549bf786d009b0/specification/resources/resource-manager/Microsoft.Authorization/stable/2018-03-01/policyAssignments.json#L37

Whenever the swagger spec specifies something generic like '{scope}' without any way of inferring what the scope can be.

generator/models.ts
anthony-c-martin: It doesn't really fit in well with any other category, and `Extension` is how it's referred to by https://github.com/Azure/autorest.azureresourceschema

Tiano2017:
there are a few conflicts left.

PR #716, opened 2019-09-24
Microsoft.Authorization policy artifacts (2019-01-01 and 2019-06-01)

pilor:
Adding 2019-01-01 and 2019-06-01 schemas for the policy entities in Microsoft.Authorization.  Also updating the special policyDefinition rule schema for both versions.

Compare schemas to iteration 1 to see changes from previous api-version (2018-05-01).
Compare tests to iteration 2 to see changes from previous api-version (2018-05-01).


pilor:
@anthony-c-martin @ifeoluwaokunoren ready for review

pilor:
BTW, I don't think this is a good candidate for auto-generation yet as the swagger specs live in the same folder as the ARM specs (resources).  I'm assuming since that isn't turned on yet that there are some complications there.

Issue #541, opened 2018-11-23
ARM Schemas contain invalid byte order marker


dowlingw:
**Background**
API specifications for `2014-04-01-preview` onwards appear to contain a byte-order marking in violation of RFC7159 S8.1:

>    Implementations MUST NOT add a byte order mark to the beginning of a
>   JSON text.  In the interests of interoperability, implementations
>   that parse JSON texts MAY ignore the presence of a byte order mark
>   rather than treating it as an error.

**What was expected**
Using third party tools to manipulate schemas from this repository to be able to open and work with JSON Schema documents without error.

**What actually happened**
When working with various JSON tools, they reject schemas from this repository as invalid JSON documents.

ifeoluwaokunoren:
Fixed with https://github.com/Azure/azure-resource-manager-schemas/pull/757

PR #757, opened 2019-10-09
Command to list all currently-defined resource types and API versions

anthony-c-martin:


generator/cmd/listresources.ts
```diff
@@ -0,0 +1,108 @@
...
+import * as constants from '../constants';
+import { series } from 'async';
+import { lowerCaseCompare } from '../utils';
+
+const rootSchemaPaths = [
+    'https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json',
+    'https://schema.management.azure.com/schemas/common/definitions.json',
+    'https://schema.management.azure.com/schemas/common/autogeneratedResources.json',
+    'https://schema.management.azure.com/schemas/common/manuallyAddedResources.json',
+];
```
ifeoluwaokunoren: Since ARM might not be the only ones consuming this command, we should add a note - maybe in the readme- so users can know the output would be only for the ResourceGroup scoped root schemas.

anthony-c-martin:
This PR inadvertently fixes https://github.com/Azure/azure-resource-manager-schemas/issues/541

PR #797, opened 2019-10-29
Add attached database configurations to schema

docohe:
In this change:

- Add new resource: Attached database configurations

- readWrite and readOnlyFollowing database.


docohe:
Please don't merge yet

anthony-c-martin:
**Summary from internal email:**

Would you guys mind testing this schema out locally to make sure the Intellisense suggestions make sense? I’ve been meaning to document this, but basically you can:
1.	Check out the repo at your branch
2.	Run “cd tools”
3.	Run “npm install”
4.	Run “npm run serve” (this’ll run a local schema server on port 3000)
5.	Open the attached template with VSCode, and replace {yourhostnamehere} with the hostname reported by step 4
6.	Ensure that there are no validation issues, and test out the autocomplete suggestions inside “properties” with Ctrl+space

Some things I noticed doing this (I’ll call these out on the PR):
1.	The Microsoft.Kusto/clusters/AttachedDatabaseConfigurations resource has only been declared inside the parent resource body, so there’s no definition to allow this resource to be deployed independently.
2.	The type has not been set correctly for being declared inside a parent resource – this should just be “AttachedDatabaseConfigurations”, as the rest of the type is inferred from the parent resource.

See the “caches_storageTargets” resource in https://github.com/Azure/azure-resource-manager-schemas/blob/master/schemas/2019-11-01/Microsoft.StorageCache.json for some examples of how to declare a resource which is deployable both independently and inside a parent resource.

docohe:
Hi,

How can we go forward with this PR

PR #831, opened 2019-12-20
Add an identity property to the HealthcareApis service

feordin:
We are adding the option to create a managed identity for the Azure API for FHIR service.  This PR adds the identity property to the service properties.


schemas/2019-09-16/Microsoft.HealthcareApis.json
```diff
@@ -63,6 +63,24 @@
...
+            "description": "Setting indicating whether the service has a managed identity associated with it.",
+            "properties": {
+                "type":{
+                    "description" : "Type of identity being specified, currently SystemAssigned and None are allowed.",
+                    "type": "string",
+                    "enum": [
+                        "SystemAssigned",
+                        "None"
+                    ],
+                    "x-ms-enum": {
```
anthony-c-martin: x-ms-enum isn't a JSON schema property

schemas/2019-09-16/Microsoft.HealthcareApis.json
```diff
@@ -63,6 +63,24 @@
           "type": "string",
           "description": "An etag associated with the resource, used for optimistic concurrency when editing it."
         },
+        "identity": {
+            "type": "object",
+            "description": "Setting indicating whether the service has a managed identity associated with it.",
+            "properties": {
```
anthony-c-martin: Missing template expression definitions. See https://github.com/Azure/azure-resource-manager-schemas/blob/b74d7fc92aa09f138448da61ac54e8eca7fb765a/schemas/2019-08-01-preview/Microsoft.OperationalInsights.json#L16-L26 for an example of how this is declared

schemas/2019-09-16/Microsoft.HealthcareApis.json
feordin: Updated per the comments to add a definition for the identity type

schemas/2019-09-16/Microsoft.HealthcareApis.json
feordin: Updated per the comments to add a definition for the identity type

PR #838, opened 2020-01-09
Regenerate SQR schema with autorest-beta

adibiton:
@anthony-c-martin can you please review?

### DIFF
--- a/schemas/2018-04-16/Microsoft.Insights.json
+++ b/schemas/2018-04-16/Microsoft.Insights.json
@@ -2,7 +2,7 @@
   "id": "https://schema.management.azure.com/schemas/2018-04-16/Microsoft.Insights.json#",
   "$schema": "http://json-schema.org/draft-04/schema#",
   "title": "Microsoft.Insights",
-  "description": "microsoft insights Resource Types",
+  "description": "Microsoft Insights Resource Types",
   "resourceDefinitions": {
     "scheduledQueryRules": {
       "type": "object",
--- a/schemas/2018-04-16/Microsoft.Insights.json
+++ b/schemas/2018-04-16/Microsoft.Insights.json
@@ -2,21 +2,11 @@
   "id": "https://schema.management.azure.com/schemas/2018-04-16/Microsoft.Insights.json#",
   "$schema": "http://json-schema.org/draft-04/schema#",
   "title": "Microsoft.Insights",
-  "description": "Microsoft Insights Resource Types",
+  "description": "microsoft insights Resource Types",
   "resourceDefinitions": {
     "scheduledQueryRules": {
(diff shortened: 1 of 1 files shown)


schemas/2018-04-16/Microsoft.Insights.json
```diff
@@ -2,21 +2,11 @@
   "id": "https://schema.management.azure.com/schemas/2018-04-16/Microsoft.Insights.json#",
   "$schema": "http://json-schema.org/draft-04/schema#",
   "title": "Microsoft.Insights",
-  "description": "Microsoft Insights Resource Types",
+  "description": "microsoft insights Resource Types",
```
AntonVasserman: Why change to lowercase?

adibiton: @anthony-c-martin I can manually change it back to Microsoft Insights - however it is the output from generating with autorest-beta (azureresourceschema version 3.0.67) - wdyt?

anthony-c-martin: @adibiton This is originating from the provider namespace in the swagger URI paths ('microsoft.insights' is lower-cased) - e.g. https://github.com/Azure/azure-rest-api-specs/blob/30c4f57844896fe41d0f179f5649760dad9101d4/specification/monitor/resource-manager/Microsoft.Insights/stable/2018-04-16/scheduledQueryRule_API.json#L272

I'd recommend you fix this for this file + other similar files in the swagger repo going forwards. For now, manually fixing the casing in this file sounds fine to me.

schemas/2018-04-16/Microsoft.Insights.json
```diff
@@ -7,16 +7,6 @@
     "scheduledQueryRules": {
```
AntonVasserman: It seems that default values for 'scheduledqueryrules_<alertName>_name' and 'components_<ai/workspaceName>_externalid' aren't generated, in @adibiton example while in the previous version it was generated.
How do we fix it?

schemas/2018-04-16/Microsoft.Insights.json
```diff
@@ -27,38 +17,49 @@
           "type": "string",
           "description": "Resource location"
         },
```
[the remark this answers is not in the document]
anthony-c-martin: The apiVersion field will always be populated by the ARM service during Export; the missing apiVersion is just a bug in the offline Export simulator tool.

schemas/2018-04-16/Microsoft.Insights.json
anthony-c-martin: Again I think this is just a difference between live & simulated - there's nothing in the schema file which can influence this.

PR #857, opened 2020-02-11
Merge autogenerated schemas

rileymckenna:
Merge 2020-01-01-preview with master.


anthony-c-martin:
@rileymckenna There are a lot of test failures because we're now missing a schema for Microsoft.Peering 2019-08-01-preview. I think that's because this version was added to the exclude list recently in https://github.com/Azure/azure-rest-api-specs/blob/master/specification/peering/resource-manager/readme.md#multi-apiprofile-support-for-autorest-v3-generators

Do you know why this change was made? It appears to have been added in this PR: https://github.com/Azure/azure-rest-api-specs/pull/8140

If this is intentional, I think you'll need to remove all the schema test for this version and remove all references to it.

PR #867, opened 2020-02-18
Auto update root schemas for subscription/tenant/managementGroup

leni-msft:


.vscode/launch.json
```diff
@@ -0,0 +1,48 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "List basepaths",
+      "runtimeArgs": ["-r", "ts-node/register"],
```
anthony-c-martin: Please could you refer to the npm commands in this file eg ,`npm run list-basepaths` rather than the underlying node commands? That way implementation could change without having to update this file.

generator/constants.ts
```diff
@@ -11,10 +11,25 @@ export const autoRestVerboseOutput = false;
...
 export const schemasBaseUri = 'https://schema.management.azure.com/schemas';
 export const schemasBasePath = path.join(generatorRoot, 'schemas');
-export const generatedSchemasPath = path.join(schemasBasePath, 'common/autogeneratedResources.json');
+export const resourceGroupRootSchema = {
+    file: path.join(schemasBasePath, 'common/autogeneratedResources.json'),
+    jsonPath: 'allOf[1].oneOf'
+};
+export const subscriptionRootSchema = {
+    file: path.join(schemasBasePath, '2018-05-01/subscriptionDeploymentTemplate.json'),
+    jsonPath: 'properties.resources.items.oneOf[0].allOf[1].oneOf'
```
anthony-c-martin: The reason that I created a separate "autogeneratedResources" file for resource group resources is because the root schemas sometimes get manual updates, and I wanted to avoid the generation process from interfering with these manual changes.

If this command is run in `generate-all` mode, IIRC it'll remove all references apart from the ones that have been regenerated. Can you check what happens if you run `generate-all`?

.gitignore
```diff
@@ -1,5 +1,4 @@
 .vs
-.vscode
```
anthony-c-martin: Please could you run all 4 of the autogeneration pipeline tasks (see https://dev.azure.com/azure/azure-resource-manager-schemas/_build) against this branch to ensure they run successfully and that the output is as expected.

.vscode/launch.json
```diff
@@ -0,0 +1,48 @@
+{
```
anthony-c-martin: Are any changes needed to the `list-resources` command? This command is used to generate https://github.com/Azure/azure-resource-manager-schemas/blob/master/generator/resources.json which is used for reporting. @ifeoluwaokunoren FYI

generator/constants.ts
leni-msft: Yes, I understand. In this PR, I ran `generate-single` for consumption, which has multi-scope resources. You can see the changes in the PR. It's working properly.

When running `generate-all`, it will only remove refs which are listed in whitelist, not all. So it should be fine. Here's the commit running `generate-all`: https://github.com/leni-msft/azure-resource-manager-schemas/commit/ec6e3a683e7eabac3d1f02a4ea3a5743dd22938a

.gitignore
leni-msft: Do you mean `Autogenerate (Batch 0~3)` ? Which branch should i specify?

BTW, as just mentioned, I executed `generate-all` locally, and attached the commit in last comment. All RP generated successfully. Do i still need run the pipeline tasks?

.vscode/launch.json
leni-msft: Just ran the command `npm run list-resources resources.json`, there's no changes after it.

leni-msft: These configurations are just to make it easier for local debugging.

.vscode/launch.json
leni-msft: These configurations are just to make it easier for local debugging. Do you think it's unnecessary?

.vscode/launch.json
leni-msft: Just ran the command `npm run list-resources resources.json`, there's no changes after it. Any more to do? @ifeoluwaokunoren

.vscode/launch.json
anthony-c-martin: I think it's great that you're including this! I was just thinking of using (for example):
```
{
  "type": "node",
  "request": "launch",
  "name": "List basepaths",
  "runtimeArgs": ["run", "list-basepaths"],
  "cwd": "${workspaceFolder}/generator",
  "args": ["npm"]
},
```

That way the implementation of the `list-basepaths` command can change, and we don't need to remember to update this file.

generator/constants.ts
anthony-c-martin: Thanks!

.gitignore
anthony-c-martin: Thanks! If you were able to run `generate-all` locally then no need to run the pipelines.

.vscode/launch.json
leni-msft: Got it, I would give it a try.

.vscode/launch.json
ifeoluwaokunoren: Ideally, we need resources.json to differentiate between the scope of the published schema, since it can either be resourcegroup, subscription, tenant, or managementgroup.  But at this time, we'd break the dependencies on the reporting pipeline, since this is a schema change for resources.json, so for now, let's merge all api-versions without differentiating the scope.

Here's what I think is the ideal structure:
{ 
"Microsoft.AAD/domainServices": [
    { 
      "resourceGroupLevel":
  [
    "2017-01-01",
    "2017-06-01",
    "2020-01-01"
  ]
    },
    {
        "subscriptionLevel":
  [
    "2019-01-01"
  ]
}
  ]
}

leni-msft: got it. I will do that.

PR #870, opened 2020-02-25
Add parameter for export storage account to Healthcareapis

LTA-Thinking:
Add support for setting the default storage account for the export operation.


schemas/2019-09-16/Microsoft.HealthcareApis.json
```diff
@@ -22,6 +22,7 @@
         "apiVersion": {
           "type": "string",
           "enum": [
+            "2018-08-20-preview",
```
anthony-c-martin: Please create a separate file for each individual API version

LTA-Thinking: Removed reference to old version

PR #1002, opened 2020-05-06
Schema for Network 2020-04-01

number213:


number213:
@anthony-c-martin could you please take a look?

PR #1041, opened 2020-07-01
Removed SQL reference due to no longer valid resources

jorgecotillo:
AutogeneratedResources.json shos that servers_backupLongTermRetentionVaults and servers_databases_backupLongTermRetentionPolicies got removed, therefore removing these references from template schemas.

This update will make the unit tests to pass again.


babbageclunk:
Hi - would you be able to push the updated version to schema.management.azure.com? We're working on a tool that generates API types based on the ARM schema, and the broken references are causing it problems. (I'm going to make it more resilient to this kind of issue as well. :)

Issue #1048, opened 2020-07-08
Inconsistencies in 2015-01-01 deploymentTemplate


matthchr:
There are a number of issues in the [2015-01-01 template](https://github.com/Azure/azure-resource-manager-schemas/blob/master/schemas/2015-01-01/deploymentTemplate.json) which I've happened to run across after running some automated tooling across it.

## Resources without `Name`
There are 128 different resources referenced which do not have a `Name` - as far as I can tell see [here](https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/resource-api-reference.md#get-resource), `Name` is a required property for the 
ARM JSON schema. Without a name, it's not actually possible to submit a resource to ARM, you get an error:

> The request content was invalid and could not be deserialized: 'Required property 'name' not found in JSON. Path 'properties.template.resources[0]', line 19, position 9.'.

Here are some examples (not all 128 of them):
1. https://schema.management.azure.com/schemas/2016-06-01/Microsoft.RecoveryServices.json#/resourceDefinitions/vaults
2. https://schema.management.azure.com/schemas/2016-06-01/Microsoft.RecoveryServices.json#/resourceDefinitions/vaultsExtendedInformation
3. https://schema.management.azure.com/schemas/2015-06-01/Microsoft.KeyVault.json#/resourceDefinitions/secrets
4. https://schema.management.azure.com/schemas/2016-08-01/Microsoft.Web.json#/resourceDefinitions/sites

I specifically tested deploying a 2015-06-01 secret into a KeyVault and confirmed that even though the schema is missing `name`, it is indeed required and the template deployment doesn't succeed without it.

There are a bunch of others spread across the following specs:
microsoft.recoveryservices/v20160601
microsoft.keyvault/v20150601
microsoft.keyvault/v20161001
microsoft.web/v20140601
microsoft.web/v20150801
microsoft.domainregistration/v20150401
microsoft.certificateregistration/v20150801
microsoft.web/v20160301
microsoft.web/v20160801
microsoft.web/v20160901
microsoft.insights/v20140401
microsoft.visualstudio/v20140226
microsoft.notificationhubs/v20150401
microsoft.network/v20151101
microsoft.network/v20170301
microsoft.network/v20170501
microsoft.storage/v20150801
microsoft.storage/v20160101
microsoft.storage/v20180201
microsoft.storage/v20180701
microsoft.storage/v20181101
microsoft.storage/v20190401
microsoft.storage/v20190601
[...]

PR #1057, opened 2020-07-17
Add missing references to 2015-01-01 main spec

matthchr:
This is related to #1048 but doesn't fix it all.


jorgecotillo:
@neilpeterson this PR LGTM, but would like a second check from you

schemas/2015-01-01/deploymentTemplate.json
```diff
@@ -2077,13 +2197,7 @@
                   { "$ref": "https://schema.management.azure.com/schemas/2019-06-01-preview/Microsoft.Synapse.json#/resourceDefinitions/workspaces_sqlPools_transparentDataEncryption" },
                   { "$ref": "https://schema.management.azure.com/schemas/2019-06-01-preview/Microsoft.Synapse.json#/resourceDefinitions/workspaces_sqlPools_vulnerabilityAssessments" },
                   { "$ref": "https://schema.management.azure.com/schemas/2019-06-01-preview/Microsoft.Synapse.json#/resourceDefinitions/workspaces_sqlPools_vulnerabilityAssessments_rules_baselines" },
-                  { "$ref": "https://schema.management.azure.com/schemas/2018-09-01-preview/Microsoft.ResourceGraph.json#/resourceDefinitions/queries" },
-                  { "$ref": "https://schema.management.azure.com/schemas/2016-07-12-preview/Microsoft.Advisor.json#/resourceDefinitions/recommendations_suppressions" },
```
neilpeterson, line 2081: Why are these being removed? May be good reason, but I'm not seeing it. I've spot checked a few of these to confirm that the schemas still exist.

matthchr: All of these URLs 404 - I can add them back if they're like supposed to be published or something (but haven't been yet?)

neilpeterson: Ah, yep, I see it. 

@jorgecotillo I do not know the publishing end of this very well. I can see that all of these have schema files in this repo. Do we need to check on the published status for _schema.management.azure.com_?

Perhaps this is a non-issue, if so LGTM.

PR #1071, opened 2020-08-07
Add missing resource reference

matthchr:


matthchr:
I filed:
https://github.com/Azure/autorest.azureresourceschema/issues/61
https://github.com/Azure/autorest.azureresourceschema/issues/62

At the very least, the changes to the `schemas/2015-01-01/deploymentTemplate.json` file aren't autogenerated though, right? Since if they were presumably they wouldn't have lost the resource reference which I added?

Also: Are these schemas really being generated? For example the commit history for:  schemas/2014-02-26/microsoft.visualstudio.json doesn't make it look like it's been generated (Since the last commit of substance was nearly 3 years ago)

matthchr:
@anthony-c-martin I've updated this PR to only touch the rollup file -- I think that this change is correct and doesn't touch any generated files (at least as I understand it).

Take a look and let me know if this looks right to you.

PR #1191, opened 2020-09-10
Add scope property to deployment template schema

filizt:
Adding "scope" property to deployment template schema. 

I wasn't able to test this locally or create a unit test for it - the PR is based on assumption that "scope" property should be recognized as a deployment template property and not a resource property. Please let me know if there is simple way to test it either locally or through unit testing.

### DIFF
--- a/schemas/2018-05-01/subscriptionDeploymentTemplate.json
+++ b/schemas/2018-05-01/subscriptionDeploymentTemplate.json
@@ -399,6 +399,10 @@
             "copy": {
               "$ref": "#/definitions/resourceCopy"
             },
+            "scope": {
+              "type": "string",
+              "description": "Scope property to be used to provide scope for a resource or to be used to specify a scope for Management Group or Tenant level deployments."
+            },
             "comments": {
               "type": "string"
             }
--- a/schemas/2019-04-01/deploymentTemplate.json
+++ b/schemas/2019-04-01/deploymentTemplate.json
@@ -1622,7 +1622,7 @@
             },
             "scope": {
               "type": "string",
-              "description": "Scope property to be used to provide scope for a resource or to be used to specify a scope for Management Group or Tenant level deployments."
+              "description": "Scope for the resource or deployment. Today, this works for two cases: 1) setting the scope for extension resources 2) deploying resources to the tenant scope in non-tenant scope deployments"
             },
             "comments": {
               "type": "string"
--- a/schemas/2019-04-01/deploymentTemplate.json
+++ b/schemas/2019-04-01/deploymentTemplate.json
@@ -1620,6 +1620,10 @@
             "copy": {
               "$ref": "#/definitions/resourceCopy"
             },
+            "scope": {
+              "type": "string",
+              "description": "Scope property to be used to provide scope for a resource or to be used to specify a scope for Management Group or Tenant level deployments."
+            },
             "comments": {
               "type": "string"
             }
--- a/schemas/2014-04-01-preview/deploymentTemplate.json
+++ b/schemas/2014-04-01-preview/deploymentTemplate.json
@@ -2144,6 +2144,10 @@
         "tags": {
           "type": "object",
           "description": "Name-value pairs to add to the resource"
+        },
+        "scope": {
+          "type": "string",
+          "description": "Scope for the resource or deployment. Today, this works for two cases: 1) setting the scope for extension resources 2) deploying resources to the tenant scope in non-tenant scope deployments"
         }
       },
       "required": [
--- a/schemas/2015-01-01/deploymentTemplate.json
+++ b/schemas/2015-01-01/deploymentTemplate.json
@@ -2335,6 +2335,10 @@
             "copy": {
               "$ref": "#/definitions/resourceCopy"
             },
+            "scope": {
+              "type": "string",
+              "description": "Scope for the resource or deployment. Today, this works for two cases: 1) setting the scope for extension resources 2) deploying resources to the tenant scope in non-tenant scope deployments"
+            },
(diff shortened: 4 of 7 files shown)


schemas/2019-04-01/deploymentTemplate.json
```diff
@@ -1620,6 +1620,10 @@
             "copy": {
               "$ref": "#/definitions/resourceCopy"
             },
+            "scope": {
+              "type": "string",
+              "description": "Scope property to be used to provide scope for a resource or to be used to specify a scope for Management Group or Tenant level deployments."
```
alex-frankel, line 1625: "to provide a scope for a resource" makes me think I can change the scope of any resource to any scope. Should we be more specific and say it works for two cases (today):
* setting the scope for extension resources
* deploying resources to the tenant scope in non-tenant scope deployments

jorgecotillo:
LGTM, only question I have is to make sure that these are the only api-versions supported, for example deploymentTemplate also exists in 2015-01-01

jorgecotillo (approved):

jorgecotillo:
@anthony-c-martin can you please review this PR as well, is adding a new property on deployment templates, looks good to me but would like your confirmation as well.

anthony-c-martin:
I think this also needs corresponding updates to:
schemas/2015-01-01/deploymentTemplate.json
schemas/2019-03-01-hybrid/deploymentTemplate.json
schemas/2014-04-01-preview/deploymentTemplate.json

schemas/2018-05-01/subscriptionDeploymentTemplate.json
```diff
@@ -399,6 +399,10 @@
             "copy": {
               "$ref": "#/definitions/resourceCopy"
             },
+            "scope": {
+              "type": "string",
+              "description": "Scope property to be used to provide scope for a resource or to be used to specify a scope for Management Group or Tenant level deployments."
```
anthony-c-martin, line 404: Do we have documentation published on how to use this? This message is going to show up in VSCode when you mouse over the property, so just want to make sure the expected usage is example; I'm not sure I'd be able to figure it out if just presented with this description.

StephenWeatherford: What effect does it have on deploymen?

StephenWeatherford:
> @StephenWeatherford just a heads-up that a new property is being added to the `resourceBase` definition. Does this have any impact on or require changes to the VSCode extension?

Thanks for asking.  @jorgecotillo I believe the validation has a hard-coded list of top-level properties, right?  You'll need to add this or else we get:
> Template validation failed: Could not find member 'scope' on object of type 'Template'. Path 'scope', line 2, position 12.
Added https://github.com/microsoft/vscode-azurearmtools/issues/967 to track.

If there are behavioral changes I might need to do other work.

jorgecotillo:
> > @StephenWeatherford just a heads-up that a new property is being added to the `resourceBase` definition. Does this have any impact on or require changes to the VSCode extension?
> 
> Thanks for asking. @jorgecotillo I believe the validation has a hard-coded list of top-level properties, right? You'll need to add this or else we get:
> 
> > Template validation failed: Could not find member 'scope' on object of type 'Template'. Path 'scope', line 2, position 12.
> > Added [microsoft/vscode-azurearmtools#967](https://github.com/microsoft/vscode-azurearmtools/issues/967) to track.
> 
> If there are behavioral changes I might need to do other work.

The validation fails because the deserialization fails to find "Scope" property, the fix is to use the latest ARM Deployments code, I have an open task to do the update this week.

schemas/2018-05-01/subscriptionDeploymentTemplate.json
alex-frankel: For the `microsoft.resources/deployments` type, it controls the target scope of the deployment. Here is an example:
https://github.com/ajf214/personal-arm-templates/blob/master/ARM-Templates/testing/nested-with-scope-property.json

It can also be used for extension resources, but I don't have an example of that handy, and I don't think it's documented yet. @tfitzmac will be working on that this week.

anthony-c-martin (approved):

schemas/2018-05-01/subscriptionDeploymentTemplate.json
StephenWeatherford: @alex-frankel It doesn't look like it should have any effect on the extension?  E.g., no effect on the syntax or semantics of resources or expressions, etc.  Correct?  thx.

alex-frankel: I think there will be false negatives if the scope property is used on resources that don't support it.  E.g. if I use it on a website or vm or other tracked resource type. @filizt / @anthony-c-martin do we have an accurate list of all the extension resources? We may need to create some sort of whitelist

anthony-c-martin: See https://github.com/Azure/azure-resource-manager-schemas/blob/master/generator/autogenlist.ts - anything marked with `ScopeType.Extension`. Currently these are completely omitted from the root schemas, so we may need to do something special to support them.

filizt: @anthony-c-martin is there any way to know if a resource type is an extension resource by using the schemas in this repo? The only way I know how to tell that is through service manifest where the routing type is defined.

anthony-c-martin: For teams not onboarded for autogeneration, I believe extension resources are not output at all. I'm actually curious what the current authoring experience looks like for extension resources...

PR #1218, opened 2020-09-21
Fix node-git error caused by nodejsv13

Descatles:


generator/package.json
```diff
@@ -12,19 +12,20 @@
...
   "devDependencies": {
-    "autorest": "^3.0.6187",
-    "@autorest/azureresourceschema": "^3.0.79",
-    "@autorest/core": "^3.0.6257",
-    "@types/async": "^3.2.0",
-    "@types/lodash": "^4.14.149",
-    "@types/node": "^13.11.0",
+    "autorest": "^3.0.6247",
+    "@autorest/azureresourceschema": "^3.0.86",
+    "@autorest/core": "^3.0.6320",
```
leni-msft, line 17: is upgrading autorest version essential?

leni-msft: if so, upgrade the version used in generator.ts as well

leni-msft (approved):

Issue #64, opened 2015-11-26
Missing auto-completion for Windows VM WinRM Listeners


pcgeek86:
The ARM JSON schemas are missing auto-completion for `http` and `https` values for the `protocols` property on a Windows Remote Management (WinRM) listener.

<img width="649" alt="screen shot 2015-11-25 at 7 30 56 pm" src="https://cloud.githubusercontent.com/assets/466713/11413840/36456968-93ab-11e5-9b96-0b81f08af068.png">

Cheers,
Trevor Sullivan
Microsoft MVP: PowerShell

wenwu449:
Thanks Trevor, we will work on some improvement on this to give better

wenwu449:
Trevor, we have fixed the issue. Please refresh your schema and check if everything is fine at your side.
Thanks again for your feedback.

pcgeek86:
@wenwu449 Thanks, can you help me understand what you mean by "refresh your schema?"

Cheers,
Trevor Sullivan

Issue #63, opened 2015-11-26
Missing auto-completion for Microsoft.Network/networkInterfaces "privateIPAllocationMethod"


pcgeek86:
I'm using Visual Studio Code 0.10.2 on Mac OS X 10.11.1 El Captian. I noticed that the `privateIPAllocationMethod` is not auto-completing for the `Microsoft.Network/networkInterfaces` Resource Type.

<img width="597" alt="screen shot 2015-11-25 at 5 04 37 pm" src="https://cloud.githubusercontent.com/assets/466713/11412200/e1efa798-9396-11e5-8ddb-36680caa4587.png">

Cheers,
Trevor Sullivan
Microsoft MVP: PowerShell

FrankSiegemund:
Thank you for reporting this issue. This has been addressed by this PR: https://github.com/Azure/azure-resource-manager-schemas/issues/63

Once the schema is refreshed and it is not reflected in VS for you, here are more instructions of how to refresh schema caches: http://blogs.msdn.com/b/webdev/archive/2014/04/11/intellisense-for-json-schema-in-the-json-editor.aspx

FrankSiegemund:
The correct link to the PR is here: https://github.com/Azure/azure-resource-manager-schemas/pull/72

wenwu449:
Trevor, we have fixed the issue. Please refresh your schema and check if everything is fine at your side.
Thanks again for your feedback.

pcgeek86:
@wenwu449 Thanks, can you help me understand what you mean by "refresh your schema?"

Cheers,
Trevor Sullivan

wenwu449:
@pcgeek86, Please check "Reload a schema" section in the link shared by Frank, (http://blogs.msdn.com/b/webdev/archive/2014/04/11/intellisense-for-json-schema-in-the-json-editor.aspx)

pcgeek86:
@wenwu449 The link you posted only talks about Visual Studio. As mentioned in the ticket, I'm using Visual Studio Code. What are the directions to reload the schema for Visual Studio Code?

Cheers,
Trevor Sulivan

philon-msft:
In VSCode, use Ctrl+Shift+P to view the list of commands and run the "Reload Window" command

pcgeek86:
@philon-msft Thanks, it is `COMMAND + SHIFT + P`, but I already knew that. :) That's helpful.

pcgeek86:
[...]

PR #1230, opened 2020-09-25
Fix autorest params

leni-msft:
Main purpose of this PR is to update schema extension version to latest, which provides the following bugfix:
* support generate multiple apiVersion schemas with readme tag [#63](https://github.com/Azure/autorest.azureresourceschema/pull/63)
* support byte format string [#64](https://github.com/Azure/autorest.azureresourceschema/pull/64)

Also update the autorest core version. In order to see the changes it may cause, the batch pipelines have been triggered on this branch.

One more thing, I tried to remove the `--pass-thru:subset-reducer`, however it appears still needed, as generating servicefabric schemas would fail without it.


anthony-c-martin:
@leni-msft please could you add an explanation for this PR to the description?

anthony-c-martin (changes requested):
I'm not sure what this PR is for - please could you add a description?

leni-msft:
> @leni-msft Lei Ni FTE please could you add an explanation for this PR to the description?

Sure, added.

anthony-c-martin (approved):

PR #1268, opened 2020-10-15
Add child resource references which were missed

matthchr:
I'm not sure if you're willing to accept this or not based on the the discussion in #1048 that some of the older specs are too big.
It looks like you just recently uploaded an update to the 2015-01-01 specification and we noticed that there are some services which had new resources added but weren't included in the rollup spec.

I think the main question we have is: Is it expected to reference an API version (such as 2019-04-01 microsoft.storage) which had new resources added, but those resources were not added to the rollup spec (although there are _other_ resources from that API version in the 2015-01-01 rollup spec).


anthony-c-martin:
Thanks for fixing this!

> I think the main question we have is: Is it expected to reference an API version (such as 2019-04-01 microsoft.storage) which had new resources added, but those resources were not added to the rollup spec (although there are _other_ resources from that API version in the 2015-01-01 rollup spec).

I think it make sense to have consistency for all resources declared by a particular provider across a given apiVersion; we just don't have any tests to assert this. I'm assuming that these resource types were initially manually added, and the schema was regenerated afterwards. We don't have any process in place to automatically update the 2015-01-01 root schema.

anthony-c-martin (approved):

matthchr:
> I think it make sense to have consistency for all resources declared by a particular provider across a given apiVersion

We're in agreement then! Totally makes sense to stop adding new apiVersions to the older specs. I was surprised that "new" resources could show up in older APIVersions but I think that may occasionally happen in reality and certainly can happen to the specs if the old spec just happened to miss some that should have been there.

PR #1307, opened 2020-11-07
Enable automerge - github action

jorgecotillo:


.github/workflows/main.yml
```diff
@@ -23,3 +23,26 @@ jobs:
       - name: Run CI tests
         run: npm test
         working-directory: ./tools
+
```
anthony-c-martin, line 26: Can we make this into a separate job? I think it would be good to keep 'build' strictly for the actual verification/build

.github/workflows/automerge.yml
```diff
@@ -0,0 +1,40 @@
+name: CI
```
anthony-c-martin: Update the name - this affects what shows up in Pull Requests IIRC

.github/workflows/automerge.yml
```diff
@@ -0,0 +1,40 @@
...
+
+on:
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
anthony-c-martin: I think you should put the "if" condition here immediately under `build:` - that way the job will show as 'skipped' for PRs it doesn't apply to. Otherwise it'll look like it always runs.

.github/workflows/automerge.yml
```diff
@@ -0,0 +1,40 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v2.1.2
+        with:
+          node-version: 10.x
+
+      - name: Check RP in Autogenlist
```
anthony-c-martin: You'll probably need a step before this to run `npm ci`

generator/cmd/findbasepath.ts
```diff
@@ -0,0 +1,27 @@
...
+
+// this function analyzes a branch name and extracts the base path, then proceeds to find it in autogenlist
+// if not found, it throws an error
+executeSynchronous(async () => {
+    if (!process.argv[2]) {
+        throw new Error("Branch name missing. This cmd expects a branch name with the following format: 'sdkAutomation/[basepath]'");
+    }
+    
+    const prName = process.argv[2];
+    let basePath = prName.replace("sdkAutomation/", "");
```
anthony-c-martin, line 12: Have you discussed this naming convention with the SDK team? Just want to make sure it's going to be robust if we're going to rely on it. What happens if there's a PR spanning multiple directories? Will they know not to change this naming convention?

anthony-c-martin (approved):

PR #1327, opened 2020-11-13
Enable resource-manager authorization schema generation

jorgecotillo:


anthony-c-martin:
Doesn't this need the other changes listed under https://github.com/Azure/azure-resource-manager-schemas/blob/master/generator/README.md#onboarding-teams-to-autogeneration-pipeline?

PR #1356, opened 2020-11-20
Make autoUpgradeMinorVersion & protectedSettings optional

50Wliu:
See https://docs.microsoft.com/en-us/azure/templates/microsoft.compute/2019-07-01/virtualmachinescalesets#virtualmachinescalesetextensionproperties-object:
autoUpgradeMinorVersion and protectedSettings are optional

Technically, all the properties are optional, but in particular I think it makes sense to not mark autoUpgradeMinorVersion & protectedSettings as required.

The CustomScriptExtension I am currently using for one of my ARM deployments only has settings and not protectedSettings, and it works fine.


50Wliu:
More specifically, I suppose the correct validation would be "`commandToExecute` must be present in exactly one of `settings` and `protectedSettings`", but a quick perusal through the JSON schema doesn't seem to reveal any easy way of enforcing that.

apclouds (approved):

50Wliu:
Thank you!

PR #1384, opened 2020-12-03
Auto-approve PR

jorgecotillo:
PR that auto-approves an incoming PR when:
a) RP is onboarded in autogeneration
b) PR is coming from a well-known user

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -61,22 +61,15 @@ jobs:
           echo "AUTOGENLIST_CHECK_RESULT=$check_result" >> $GITHUB_ENV;
         working-directory: ./generator
         
-      - name : Apply Merge Label
+      - uses: hmarr/auto-approve-action@v2.0.0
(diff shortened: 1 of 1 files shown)


anthony-c-martin:
@jorgecotillo What's the motivation for this change? FWIW I did some digging into the errors you're getting when the pipeline runs:
```
2020-12-02T08:21:28.295Z INFO  Current PR status: mergeable_state: blocked
2020-12-02T08:21:28.295Z INFO  Retrying after 5000 ms ... (1/6)
2020-12-02T08:21:33.576Z INFO  Current PR status: mergeable_state: blocked
2020-12-02T08:21:33.576Z INFO  Retrying after 5000 ms ... (2/6)
...
```

I believe these errors are coming back because the (undocumented?) mergeable_state field in this API is returning 'blocked': https://developer.github.com/v3/pulls/#get-a-pull-request - see the source for this action here: 
https://github.com/pascalgn/automerge-action/blob/5664da29c4edb3b5c8cfd3c950c23fcac92696d0/lib/merge.js#L227-L239

My assumption is that this is because of this:
image

PR #1402, opened 2020-12-08
Sdk auto/containerregistry

leni-msft:
replaces https://github.com/Azure/azure-resource-manager-schemas/pull/1349


leni-msft:
After resolving conflicts and rerun `generate-single`, it turns out to be no changes... So I close this PR.

PR #1446, opened 2021-01-04
Replicate changes from #824 for other-scoped templates

### LINKED ISSUE #1064: 2018-05-01/subscriptionDeploymentTemplate.json requires output property "value"
When producing a dynamic list of outputs, an output property of 'value' is not required - [docs](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-outputs?tabs=azure-powershell#dynamic-number-of-outputs).

This is confirmed when looking at [2019-04-01/deploymentTemplate.json](https://github.com/Azure/azure-resource-manager-schemas/blob/master/schemas/2019-04-01/deploymentTemplate.json#L1886).

However when looking at [2018-05-01/subscriptionDeploymentTemplate.json](https://github.com/Azure/azure-resource-manager-schemas/blob/master/schemas/2018-05-01/subscriptionDeploymentTemplate.json#L620), this property is being enforced. 

This is causing a validation issue when using the ARM Tools VS Code extension. Extension issue with example can be found here https://github.com/microsoft/vscode-azurearmtools/issues/695#issuecomment-663235893 .

Thanks

anthony-c-martin:
Closes #1064

### DIFF
--- a/schemas/2018-05-01/subscriptionDeploymentTemplate.json
+++ b/schemas/2018-05-01/subscriptionDeploymentTemplate.json
@@ -649,6 +649,17 @@
     "output": {
       "type": "object",
       "properties": {
+        "condition": {
+          "oneOf": [
+            {
+              "type": "boolean"
+            },
+            {
+              "$ref": "https://schema.management.azure.com/schemas/common/definitions.json#/definitions/expression"
+            }
+          ],
+          "description": "Condition of the output"
+        },
         "type": {
           "$ref": "#/definitions/parameterTypes",
           "description": "Type of output value"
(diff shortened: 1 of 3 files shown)


schemas/2018-05-01/subscriptionDeploymentTemplate.json
```diff
@@ -649,18 +649,32 @@
...
+        },
+        "copy": {
+          "$ref": "#/definitions/outputCopy",
+          "description": "Output copy"
         }
       },
       "required": [
-        "type",
-        "value"
+        "type"
```
jorgecotillo, line 677: just curious, is value no longer required?

jorgecotillo (approved):

schemas/2018-05-01/subscriptionDeploymentTemplate.json
anthony-c-martin: It's not required when using "copy" - see [here](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-outputs?tabs=azure-powershell#dynamic-number-of-outputs) for an example.

PR #1485, opened 2021-01-29
Add quantum to older deployment schemas

xinyi-joffre:
Adding Azure Quantum to the deployment template schemas for 2014-04-01-preview and 2015-01-01.


xinyi-joffre:
@anthony-c-martin, I think this was the missing part from the onboarding PR. After completing the last PR, we still were not able to export template for Azure Quantum Workspaces--it looked like the portal was using deployment template schema from 2015-01-01.

anthony-c-martin:
Happy to merge this, but note that the ARM service does use 2019-04-01 as the source of resource schema definitions to export.

anthony-c-martin (approved):

xinyi-joffre:
Thank you, @anthony-c-martin! It looks like it is still not working for us:

Screen Shot 2021-02-01 at 8 36 49 AM

Do you have any idea what might be going wrong?

anthony-c-martin:
> Do you have any idea what might be going wrong?

The published schemas must be picked up and integrated into the next ARM build, so there's a delay between checking schemas into this repo and having them be available for Export. Feel free to ping me internally (antmarti) if you'd like an estimate for when the schemas will be picked up for Export.

PR #1518, opened 2021-02-22
Fix missing references in 2015-01-01 schema

matthchr:
I _think_ there was agreement a while back that if an API version is referred to by the 2015 spec, it should refer to all of the entities in that API version (not just some). This resolves a few that were missed based on the recent updates to the schema that were published last week.


matthchr:
@anthony-c-martin can you take a look at this when you get a chance?

anthony-c-martin (approved):

anthony-c-martin:
Thanks for fixing this up!

PR #1521, opened 2021-02-23
Enable schema loader test from Deployments Templates

jorgecotillo:
This PR loads all schemas, using current Azure.Deployments.Templates and verifies that all can be loaded successfully. 
Common error this test detects is duplicate properties such as "type".


tools/deploymentsSchemaTests/DeploymentsTests.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace deploymentsSchemaTests
+{
+    [TestClass]
+    public class DeploymentsTests
+    {
+        [TestMethod]
+        public void TestSchemaLoader()
+        {
+            // Calling Path.Combine to construct the path to schemas folder
+            var schemasPath = Path.Combine(new string[] { "..", "..", "..", "..", "..", "schemas" });
```
jorgecotillo, line 15: >new string[] { "..", "..", "..", "..", "..", "schemas" } [](start = 43, length = 56)

@anthony-c-martin is there a better way to construct a path? relying on the number of parents seems a little fragile, although, is unlikely to change ...

tools/deploymentsSchemaTests/TestSchemaCache.cs
```diff
@@ -0,0 +1,166 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace deploymentsSchemaTests
+{
+    /// <summary>
+    /// Test schema cache
+    /// </summary>
+    internal class TestSchemaCache : INormalizedSchemaCache
```
jorgecotillo, line 21: >TestSchemaCache [](start = 19, length = 15)

Had to create a custom class, for the most part this code is identical to the OfflineSchemaCache. When I tried to use CreateForNormalizedSchemas (and passed all schemas folders) I got an error saying that some schemas (37 actually) were not found, some of them had to do with CreateUIDefinition.CommonControl.json (the ones from 0.0.1-preview and 0.X.X-preview), so my approach was to use the same code CreateFromAssembly implements with the difference that instead of loading schemas from the assembly I use filePaths instead.

tools/deploymentsSchemaTests/DeploymentsTests.cs
anthony-c-martin: IMO declaring the dependency in the .csproj is more robust - e.g. something like:
```
<ItemGroup>
  <None Include="../../schemas/**" CopyToOutputDirectory="PreserveNewest" />
</ItemGroup>
```

That'll copy the files to the output directory at build time, so they should end up in the same dir as the DLLs, and you can rely on the path to reach them not changing. It should also fail the build if the path is incorrect.

anthony-c-martin: Note - you may also need to add a `[DeploymentItem]` attribute to your tests to register the schema files as test dependencies.

tools/deploymentsSchemaTests/TestSchemaCache.cs
```diff
@@ -0,0 +1,166 @@
...
+using Azure.Deployments.Templates.Helpers;
+using Azure.Deployments.TemplateSchemas;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace deploymentsSchemaTests
```
anthony-c-martin, line 16: I think we should stick to the .NET convention of pascal-case for namespace / assembly names. ("deploymentsSchemaTests" -> "DeploymentsSchemaTests")

tools/deploymentsSchemaTests/deploymentsSchemaTests.csproj
```diff
@@ -0,0 +1,18 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Azure.Deployments.Templates" Version="1.0.154" />
```
anthony-c-martin, line 10: This reminds me - we should probably onboard this repo for dependabot, so that we pick up updates automatically. Here's how we have it configured for Bicep: https://github.com/Azure/bicep/blob/main/.github/dependabot.yml. I'll create an issue for this.

anthony-c-martin (approved):
Nice one! This will be great to have

PR #1554, opened 2021-03-19
Added B1ls sku for autocomplete

johnlokerse:
The autocomplete for the B1ls virtualmachine sku is missing. I added reference to the Microsoft.Compute.json.

Microsoft documentation: https://docs.microsoft.com/en-us/azure/virtual-machines/sizes-b-series-burstable


anthony-c-martin:
@grizzlytheodore please could you take a look?
1. Do we need to make this same change to the other Microsoft.Compute schemas?
1. Do we need to make this change to Microsoft.Compute swagger specs?
1. How do we ensure this change gets picked up by future schema checkins?

PR #1581, opened 2021-04-15
Remove AML Compute FileServer

wzhang373:


JeffreyRichter (approved):

itsmona25 (approved):

anthony-c-martin:
@wzhang373 Your team is onboarded for [automatic schema generation](https://github.com/Azure/azure-resource-manager-schemas/tree/master/generator#onboarding-teams-to-autogeneration-pipeline) - meaning you don't need to manually submit changes to this repo.

The change being referenced (https://github.com/Azure/azure-rest-api-specs/pull/13944) does not remove the PUT definition for this resource type, so it is still accepted on a template deployment - as such schemas will continue to declare support for it. Do you intend to deprecate the PUT?

wzhang373:
My intention is to deprecate all APIs, except List calls. If I remove PUT definition in spec, autorest will not generate the List APIs.  This is the reason I keep the PUT api but remove the declaration in ARM deployment template.

Let me know if you have better way to not declare support for PUT but still leave List APIs in place.

anthony-c-martin:
As discussed offline, schema generation is driven by what has been declared in the azure-rest-api-specs repo. Once the appropriate PUT APIs have been removed there, this change will be automatically made here.

PR #1756, opened 2021-06-11
Add suffix to Microsoft.Authorization config

anthony-c-martin:
This fixes a break in the autogeneration pipeline


jorgecotillo (approved):

matsest:
@anthony-c-martin  @jorgecotillo  FYI, we had a reference in our scripts to this published schema file https://schema.management.azure.com/schemas/2019-09-01/Microsoft.Authorization.json#/definitions/PolicyDefinitionReference which is now gone, and breaks our validations.

Works fine after changing it to `.Resources.json,` but still - this is a breaking change.

Why not keep both blobs published at schema.management.azure.com?

anthony-c-martin:
@matsest - we've been working off the assumption that external schema validation would use the [root schema](https://github.com/Azure/azure-resource-manager-schemas/blob/master/schemas/2019-04-01/deploymentTemplate.json) as an entry point, and that structural changes are OK as long as they don't affect the semantics.

Would you mind sharing a little more detail about your use case so that we can try and avoid breaking you in future?

matsest:
So what we do now is approx:
- Deploying Azure Policy with custom content (policies, initiatives and assignments) through New-AzPolicyDefinition in a script
- To have useful json validation both when authoring and pre-deployment we need some sort of validation based on a schema, as authoring Azure Policy is - sadly - not the most straightforward thing to do.

What do we do? The same thing as here: https://github.com/Azure/azure-resource-manager-schemas/blob/deb9788202e349cc8f47dba93b1c49dbaeb8967c/tests/2019-01-01/Microsoft.Authorization.PolicySetDefinitions.tests.json#L5 

We run validation tests with Json validation. Hence we point to the specific schema for this:

      "definition": "https://schema.management.azure.com/schemas/2019-01-01/Microsoft.Authorization.Resources.json#subscription_resourceDefinitions/policySetDefinitions",<br class="Apple-interchange-newline">

Where there schema address now changed.

If there's a better entry point for this that would be great.

anthony-c-martin:
@matsest Thanks for the explanation. We do have logic [here](https://github.com/Azure/azure-resource-manager-schemas/blob/master/generator/cmd/listresources.ts) to build a list of available resources which you could modify to spit out $ref paths, but it feels unnecessarily complex for your scenario, and prone to similar brittle-ness. It's used to generate [this list](https://github.com/Azure/azure-resource-manager-schemas/blob/master/generator/resources.json).

Now that policy have been onboarded for automatic generation, I think it's unlikely that this file path will change again - but difficult to say it's impossible. A cheap option might be to just contribute a unit test to this repo which validates the file exists at the path and prints out a message like "Please try to avoid changing this path, and let @matsest know in advance if it must be changed.".

PR #1761, opened 2021-06-14
Update deploymentTemplate schema files

lwang2016:
Updated deploymentTemplate schema files to support resource symbolic names. Added the "SymbolicName" version of template tests to cover the new scenario.


schemas/2014-04-01-preview/deploymentTemplate.json
```diff
@@ -2461,6 +370,2119 @@
...
+                { "$ref": "https://schema.management.azure.com/schemas/2020-05-01/Microsoft.Network.json#/resourceDefinitions/virtualNetworks_subnets" },
+                { "$ref": "https://schema.management.azure.com/schemas/2020-05-01/Microsoft.Network.json#/resourceDefinitions/virtualNetworks_virtualNetworkPeerings" },
+                { "$ref": "https://schema.management.azure.com/schemas/2020-05-01/Microsoft.Network.json#/resourceDefinitions/virtualNetworkTaps" },
+                { "$ref": "https://schema.management.azure.com/schemas/2020-05-01/Microsoft.Network.json#/resourceDefinitions/virtualRouters" },
+                { "$ref": "https://schema.management.azure.com/schemas/2020-05-01/Microsoft.Network.json#/resourceDefinitions/virtualRouters_peerings" },
+                { "$ref": "https://schema.management.azure.com/schemas/2020-05-01/Microsoft.Network.json#/resourceDefinitions/virtualWans" },
+                { "$ref": "https://schema.management.azure.com/schemas/2020-05-01/Microsoft.Network.json#/resourceDefinitions/vpnGateways" },
+                { "$ref
```
anthony-c-martin, line 2485: This should be `items` rather than `additionalProperties` to declare the schema for array contents.

anthony-c-martin, line 2475: I think it would be worth trying to come up with a more descriptive name here - as this description will be shown in IDE hovers & validation.

anthony-c-martin, line 2482: I think it would be worth trying to come up with a more descriptive name here - as this description will be shown in IDE hovers & validation.

schemas/2015-01-01/deploymentTemplate.json
```diff
@@ -48,2267 +48,15 @@
       "description": "User defined functions"
     },
     "resources": {
```
anthony-c-martin: I take it you've passed the structural changes by Stephen? Are changes to the ARM Tools extension needed before this PR is merged?

schemas/2015-01-01/deploymentTemplate.json
```diff
@@ -2756,6 +504,2280 @@
         "input"
       ],
       "description": "Output copy"
+    },
+    "resourceSchemaBase": {
```
anthony-c-martin: Could we just call this "resource"?

anthony-c-martin:
It looks like the structural changes in this PR need to be accounted for in the autogenerator code as it relies on being able to programmatically update the list of resources. I get an exception when I run the following:

```sh
cd generator/
npm ci
npm run generate-single -- policyinsights/resource-manager
```

You should just need to make minor changes here to account for the changing jpath:
https://github.com/Azure/azure-resource-manager-schemas/blob/b23f790a5dddf24f8a6420b1eb9276890e3beca3/generator/constants.ts#L16-L31

anthony-c-martin (changes requested):
Needs a few small changes, but looks good!

schemas/2014-04-01-preview/deploymentTemplate.json
lwang2016: Good catch. Fixed.

PR #1767, opened 2021-06-17
Specify full path for autorest invokation

anthony-c-martin:
I noticed that on more recent versions of `node`, the path resolution logic for `autorest.cmd` no longer works.

This should allow it to work in a back-compatible manner.

### DIFF
--- a/generator/generate.ts
+++ b/generator/generate.ts
@@ -100,7 +100,7 @@ async function handleGeneratedSchema(readme: string, schemaPath: string, autoGen
 }
 
 async function execAutoRest(tmpFolder: string, params: string[]) {
-    await executeCmd(__dirname, autorestBinary, params);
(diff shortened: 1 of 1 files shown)


generator/generate.ts
```diff
@@ -100,7 +100,7 @@ async function handleGeneratedSchema(readme: string, schemaPath: string, autoGen
 }
 
 async function execAutoRest(tmpFolder: string, params: string[]) {
-    await executeCmd(__dirname, autorestBinary, params);
+    await executeCmd(__dirname, `${__dirname}/node_modules/.bin/${autorestBinary}`, params);
```
jorgecotillo, line 103: >${__dirname}/node_modules/.bin/${autorestBinary} [](http://example.com/codeflow?start=33&length=48)

I remember having issues in the past with platform specific path delimiters, not sure if makes sense to use [path.join](https://nodejs.org/api/path.html#path_path_join_paths)

jorgecotillo (approved):

PR #1779, opened 2021-06-25
Fix a typo in deploymentTemplate.json 2015-01-01

lwang2016:
Fix a typo in deploymentTemplate.json 2015-01-01


anthony-c-martin (approved):
Thanks for fixing this!

PR #1782, opened 2021-07-01
Autogenerate

anthony-c-martin:


jorgecotillo (approved):

generator/processors/Microsoft.SecurityInsights.ts
```diff
@@ -0,0 +1,6 @@
+import { SchemaPostProcessor } from '../models';
+import { replaceCyclicRef } from './helpers';
+
+export const postProcessor: SchemaPostProcessor = (namespace: string, apiVersion: string, schema: any) => {
+  replaceCyclicRef(schema.definitions?.MetadataDependencies?.properties?.criteria?.oneOf[0]?.items);
```
jorgecotillo, line 5: Pretty cool, this way we can easily update cyclic references, also I think it is worth discussing how we can enhance the current processor to handle these type of references.

PR #1793, opened 2021-07-07
Update deploymentTemplate schema files to support resource symbolic name

lwang2016:
Update deploymentTemplate schema files to support resource symbolic name


anthony-c-martin (approved):
Looks good!

PR #1794, opened 2021-07-07
Add evaluationDelay to policyDefinition rule schema

pilor:
ifNotExists policy rules now support a configurable evaluation delay. It can't be a string enum (without a bunch of crazy logic) because it can either be a few set values ("AfterProvisioning[Success|Failure]") or an ISO duration (i.e. "PT2H").

1st iteration copied existing versions to new paths, 2nd+ iterations have changes.


pilor:
@anthony-c-martin policy rule schema update.  1st iteration is a copy of the old schema, 2nd is the minor update.  All tests passing

anthony-c-martin (approved):

anthony-c-martin:
Looks good!

PR #1820, opened 2021-08-04
Onboard new RPs for autogeneration by default

anthony-c-martin:


generator/autogenlist.ts
```diff
@@ -8,8 +8,224 @@ import { postProcessor as policyProcessor } from './processors/Microsoft.Authori
...
+        basePath: 'videoanalyzer/resource-manager',
+        namespace: 'Microsoft.Media',
+        disabledForAutogen: true,
+    },
+    {
+        basePath: 'webpubsub/resource-manager',
+        namespace: 'Microsoft.SignalRService',
+        disabledForAutogen: true,
+    },
+];
```
anthony-c-martin, line 224: This list of a snapshot of the `list-basepaths` output - to avoid changing the list of RPs currently opted in/out.

apclouds (approved):

generator/autogenlist.ts
```diff
@@ -268,10 +484,10 @@ const autoGenList: AutoGenConfig[] = [
         basePath: 'EnterpriseKnowledgeGraph/resource-manager',
         namespace: 'Microsoft.EnterpriseKnowledgeGraph',
     },
-    /*{ NOTE(jcotillo): Temporally removing this RP - latest swagger contains an unsupported type by the schema generator tool
```
jorgecotillo, line 271: >latest swagger contains an unsupported type [](http://example.com/codeflow?start=54&length=43)

I recall we had issues when generating this schema, would this break our schema generation?

jorgecotillo:
> export const blocklist = [

Something I quite didn't get is what is the criteria to decide which RPs go here, which ones don't, do we just wait until the schema generation fails and we blocklist them?

---
Refers to: generator/constants.ts:38 in d1f8696. [](commit_id = d1f8696736f3fa1c9a7ee9d63b42d3306a353a0d, deletion_comment = False)

generator/autogenlist.ts
```diff
@@ -8,8 +8,224 @@ import { postProcessor as policyProcessor } from './processors/Microsoft.Authori
 import { postProcessor as securityInsightsPostProcessor } from './processors/Microsoft.SecurityInsights';
 import { lowerCaseEquals } from './utils';
 
+// Going forwards, providers are onboarded by default.
+// The providers listed here are the only ones **not** onboarded.
+const disabledProviders: AutoGenConfig[] = [
```
jorgecotillo, line 13: >disabledProviders [](http://example.com/codeflow?start=6&length=17)

What is the difference between this and blocklist?

anthony-c-martin: Good question. I see the blocklist more like "these paths should be totally ignored, and do not apply for schema gen", whereas the config here is "these RPs have not yet been onboarded, but will be in the future". Let me add some comments and possible rename variables to try and make this clearer.

PR #1825, opened 2021-08-06
Remove cloudServices/updateDomains resource from ARM Schema

sanchitkum:
The ARM Schemas defined here is used during Export Template of a resource. In the cloud services resource, we don't need the update domain resource to be exported by default, hence removing the same from the schema.


jorgecotillo:
LGTM, @anthony-c-martin for final confirmation.

anthony-c-martin (approved):

PR #1859, opened 2021-08-25
add compute 2021-07-01 and 2021-04-01

grizzlytheodore:


grizzlytheodore:
is this part still a thing?: 
Ensure that any $refs to resource types that you are adding has been added to the following top-level template schema: schemas/2019-04-01/deploymentTemplate.json

I don't see any change in that file

anthony-c-martin:
> @anthony-c-martin
> is this part still a thing?
> `Ensure that any $refs to resource types that you are adding has been added to the following top-level template schema: schemas/2019-04-01/deploymentTemplate.json`
> 
> I don't see any change in that file from the generation

Yes, that's still required. The top-level schema is the index for discovering resource schemas, so by not adding to this file, your schemas won't be discoverable.

We've optimized the generation process for provider which are onboarded for auto generation; some manual steps are still required for RPs that are not onboarded.

PR #1863, opened 2021-08-25
Microsoft.insights monitor 2021 07 01 plscopes

arolshan:


arolshan:
Hi, after I merge the swagger, how many days does it take to auto-generate the new schemas?

anthony-c-martin:
> Hi, after I merge the swagger, how many days does it take to auto-generate the new schemas?

The process runs every 24h, so at most - 24h. You can view a summary of the results here: https://github.com/Azure/azure-resource-manager-schemas/blob/autogenerate/generator/resources.json

arolshan:
THANKS

PR #1906, opened 2021-09-14
Adding the schema for uiFormDefinition. 

ShrishaKumar:
A uiFormDefinition lets you customize the UI in the Azure portal when deploying an ARM template using the custom deployment blade


schemas/2021-09-09/uiFormDefinition.schema.json
```diff
@@ -0,0 +1,3933 @@
...
+        },
+        {
+          "$ref": "#/definitions/formStepWithoutResourceScope"
+        },
+        {
+          "$ref": "#/definitions/formStepWithoutResourceScope"
+        },
+        {
+          "$ref": "#/definitions/formStepWithoutResourceScope"
+        }
```
anthony-c-martin, line 287: I don't normally review UI schemas, but just happened to notice this, which seems odd. Do you know what's going on with the duplication here?

anthony-c-martin (approved):

PR #2003, opened 2021-10-30
Add linter validation

anthony-c-martin:
* Add linters for node projects
* Add dependabot config
* Switch to `yargs` for CLI arguments


azure-pipelines-autogen.yml
```diff
@@ -62,7 +62,7 @@ jobs:
   - script: |
       set -Eeuxo pipefail
       cd generator
-      npm run generate-all $(generateParams)
+      npm run generate-all -- --batch-count $(batchCount) --batch-index $(batchIndex)
```
jorgecotillo, line 65: >-- [](http://example.com/codeflow?start=27&length=2)

is the additional `--` intentional?

README.md
```diff
@@ -24,7 +24,7 @@ You can use the generator in this repo to automatically generate a schema from a
 ```bash
 cd generator
 npm install
-npm run generate-single myprovider/resource-manager
+npm run generate-single -- --base-path myprovider/resource-manager
```
jorgecotillo, line 27: >-- [](http://example.com/codeflow?start=24&length=2)

same here, is the additional `--` intentional?

.github/workflows/generate-types.yml
```diff
@@ -35,9 +35,9 @@ jobs:
         run: |
           if [ -z "${{ github.event.inputs.single_path }}" ]
           then
-            npm run generate-all -- '{"localPath": "$GITHUB_WORKSPACE/workflow-temp/azure-rest-api-specs"}'
+            npm run generate-all -- --local-path "$GITHUB_WORKSPACE/workflow-temp/azure-rest-api-specs"
```
jorgecotillo, line 38: >-- [](http://example.com/codeflow?start=33&length=2)

same here, if not intentional please remove.

azure-pipelines-autogen.yml
anthony-c-martin: Yup, take a look at https://stackoverflow.com/questions/43046885/what-does-do-when-running-an-npm-command. Without it, the `--<param>` args are passed to `npm` rather than to the script that `npm` is executing.

README.md
jorgecotillo: Seems like it is intentional.

.github/workflows/generate-types.yml
jorgecotillo: Seems like it is intentional.

azure-pipelines-autogen.yml
jorgecotillo: Got it, thanks for sharing.

jorgecotillo (approved):

PR #2049, opened 2021-11-12
GH actions single schema gen fix

jorgecotillo:
GH actions complained about 'output-path' not being recognized, it was due a missing '--' as part of the npm cmd.


jorgecotillo:
Chatted offline, not the real issue ... will fix it in a separate PR.

PR #2050, opened 2021-11-12
Updated GH actions argument name

jorgecotillo:
Updated argument name to allow single schema generation


anthony-c-martin (approved):
:shipit:

PR #2376, opened 2022-05-17
[Microsoft.Media] Fix tests

jorgecotillo:


tools/templateTests/Template_2.2019-04-01.json
```diff
@@ -433,20 +433,6 @@
         "additional": "value"
       }
     },
-    {
-      "type": "Microsoft.Media/mediaServices",
```
jorgecotillo, line 437: Removed these tests because it was causing some cryptic message saying that expected array doesn't match the result, not sure which array is referring to ...

PR #2431, opened 2022-06-17
Added Ms.Network to schema autogeneration list

MikhailTryakhov:


generator/autogenlist.ts
```diff
@@ -1028,6 +1023,10 @@ const autoGenList: AutoGenConfig[] = [
         namespace: 'Microsoft.Media',
         postProcessor: mediaPostProcessor
     },
+    {
+        basePath: 'network/resource-manager',
+        namespace: 'Microsoft.Network'
```
anthony-c-martin, line 1028: There are quite a few other directories in the specs repo which contribute to `Microsoft.Network` schemas (search in this file to see others). In order to avoid them conflicting with each other, we will need to add a suffix to the schema file to disambiguate. To see an example of this config, look at https://github.com/Azure/azure-resource-manager-schemas/blob/f80af50c7ec9b649d2e3d97008276fe74fd9dd7e/generator/autogenlist.ts#L503-L512.

For the other specs paths listed in this file, do you know if there is any plan to onboard?
```
dns/resource-manager
privatedns/resource-manager
trafficmanager/resource-manager
frontdoor/resource-manager
```

anthony-c-martin (dismissed):
See comments

generator/autogenlist.ts
MikhailTryakhov: yeah, they're related to that, but have no idea about their plans... we're not maintaining them

jorgecotillo: Please make sure to add a suffix as mentioned by Anthony.

jorgecotillo (approved):
LGTM!

PR #2434, opened 2022-06-20
delete childResource and cloudservices.properties.resources

grizzlytheodore:
delete childResource and cloudservices.properties.resources

### DIFF
--- a/generator/processors/Microsoft.Compute.ts
+++ b/generator/processors/Microsoft.Compute.ts
@@ -57,7 +57,7 @@ export const postProcessor: SchemaPostProcessor = async (namespace, apiVersion, 
     }
 
     /// Remove CloudService Update Domain
-    if (schema.resourceDefinitinos?.cloudServices_updateDomains) {
+    if (schema.resourceDefinitions?.cloudServices_updateDomains) {
         delete schema.resourceDefinitions.cloudServices_updateDomains;
     }
     if (schema.resourceDefinitions?.cloudServices?.resources) {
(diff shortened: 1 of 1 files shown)


generator/processors/Microsoft.Compute.ts
```diff
@@ -60,9 +60,12 @@ export const postProcessor: SchemaPostProcessor = async (namespace, apiVersion,
     if (schema.resourceDefinitions?.cloudServices_updateDomains) {
         delete schema.resourceDefinitions.cloudServices_updateDomains;
     }
-    if (schema.resourceDefinitions?.cloudServices?.resources) {
+    if (schema.resourceDefinitions?.cloudServices?.properties?.resources) {
         delete schema.resourceDefinitions.cloudServices.resources
```
anthony-c-martin, line 64: Missing `.properties` on this line

grizzlytheodore: added. thank you

generator/processors/Microsoft.Compute.ts
```diff
@@ -60,9 +60,12 @@ export const postProcessor: SchemaPostProcessor = async (namespace, apiVersion,
...
     }
+    if (schema.cloudServices_updateDomains_childResource) {
+        delete schema.cloudServices_updateDomains_childResource
+    }
```
anthony-c-martin, line 68: I believe this should be `schema.definitions?.cloudServices_updateDomains_childResource`

Note that if you're using VSCode, you can find the JSON path of the cursor by looking at the breadcrumbs at the top of the editor:
image

anthony-c-martin (approved):

generator/processors/Microsoft.Compute.ts
anthony-c-martin: No worries. I've merged. I suggest just keeping an eye on the next generation pipeline run (should be ~24h from now) and making sure it has the intended effect

Issue #2462, opened 2022-07-05
Deleted versions of Microsoft.Authorization?


matthchr:
It looks like (7a704f59e734099cef4a7fba901cca4fb0a5224d)[https://github.com/Azure/azure-resource-manager-schemas/commit/7a704f59e734099cef4a7fba901cca4fb0a5224d] removed a bunch of Microsoft.Authorization preview versions from the autogenerated JSON.

As far as I can tell, these versions are still supported and still exist in Swagger.

Was this removal intentional?

matthchr:
FWIW I just confirmed that at least 2020-08-01-preview works.

jeskew:
https://github.com/Azure/bicep/issues/6610 was caused by types missing from the provider Swagger, so I don't believe it is related to the issue reported.

jorgecotillo:
This is issue happens because the spec contains duplicate paths, the spec specifies a subscription scope path:

https://github.com/Azure/azure-rest-api-specs/blob/7f80a452b284a2cbc57c65ecc6e8224360d72657/specification/authorization/resource-manager/Microsoft.Authorization/preview/2021-12-01-preview/authorization-AccessReviewCalls.json#L66

But also include a {scope} parameter, {scope} parameter supports tenant, mg, subscription and resource group scopes, therefore subscription scope is included twice:

https://github.com/Azure/azure-rest-api-specs/blob/7f80a452b284a2cbc57c65ecc6e8224360d72657/specification/authorization/resource-manager/Microsoft.Authorization/preview/2021-12-01-preview/authorization-AccessReviewCalls.json#L998

Need to reach out to Microsoft.Authorization RP and ask them to either remove their subscription path or {scope} path.

@stephaniezyen do you have a contact to reach out to please? Seems like the PR owner is OOF until end of July,

jorgecotillo:
Latest update, will need to re-run the schema generation, seems like we had a long running bug that caused duplicated child: https://github.com/Azure/autorest.azureresourceschema/pull/81 it has been fixed, will see if this extension update fixes Microsoft.Authorization generation issue.

matthchr:
I think this is now fixed and can probably be closed? At least, I updated to use the latest from `main` and didn't have anything get deleted (like I had before)

PR #2507, opened 2022-08-02
Temporarily disable Authorization/Resource-Manager from autogen

jorgecotillo:
Workaround to handle: https://github.com/Azure/azure-resource-manager-schemas/issues/2462

The following swagger spec PR introduced duplicate paths: https://github.com/Azure/azure-rest-api-specs/pull/16919 and as a result the schema generation failed, since the schema generation process first removes all existing schemas (to simplify handling deltas between specs) and proceeds to re-generate them, failing to re-generate schemas means that an RP will end up with existing schemas getting deleted.

### DIFF
--- a/generator/autogenlist.ts
+++ b/generator/autogenlist.ts
@@ -89,6 +89,7 @@ const disabledProviders: AutoGenConfig[] = [
         namespace: 'Microsoft.Media',
         disabledForAutogen: true,
     },
+    // Disabled because the swagger spec contains duplicate API paths and results in schema generation failures, see here for more info: https://github.com/Azure/azure-resource-manager-schemas/issues/2462
     {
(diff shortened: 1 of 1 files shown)


generator/autogenlist.ts
```diff
@@ -89,6 +89,11 @@ const disabledProviders: AutoGenConfig[] = [
         namespace: 'Microsoft.Media',
         disabledForAutogen: true,
     },
+    {
+        basePath: 'authorization/resource-manager',
```
anthony-c-martin, line 93: Mind adding a comment here to explain why it's been removed?

anthony-c-martin (approved):

generator/autogenlist.ts
jorgecotillo: Comment added

PR #2835, opened 2023-03-13
Added statements to log time taken to generate schemas for each RP

Xynoclafe:
Added statements to log time taken to generate schemas for each RP to debug which RPs are causing timeouts with batches 2 and 3 of autogenerate.

### DIFF
--- a/generator/cmd/generateall.ts
+++ b/generator/cmd/generateall.ts
@@ -94,7 +94,7 @@ executeSynchronous(async () => {
                     const startTime = Date.now();
                     const newConfigs = await generateSchemas(readme, autoGenConfig);
                     const generationTime = Date.now() - startTime;
-                    console.log(`Time taken to generate ${autoGenConfig.basePath} : ${generationTime}.`)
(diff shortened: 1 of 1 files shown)


generator/cmd/generateall.ts
```diff
@@ -91,7 +91,10 @@ executeSynchronous(async () => {
                     const readme = validateAndReturnReadmePath(localPath, autoGenConfig.readmeFile || autoGenConfig.basePath);
                     pkg.packageName = getPackageString(readme);
 
+                    const startTime = Date.now();
                     const newConfigs = await generateSchemas(readme, autoGenConfig);
+                    const generationTime = Date.now() - startTime;
+                    console.log(`Time taken to generate ${autoGenConfig.basePath} : ${generationTime} ms.`);
```
jorgecotillo, line 97: To differentiate the message, consider adding colors, like here: https://github.com/Azure/azure-resource-manager-schemas/blob/3134761b9f051d366342f4f9fcc0b36fc03f6259/generator/cmd/listbasepaths.ts#L24

jorgecotillo (approved):

PR #2964, opened 2023-09-08
[initScript.sh] Use npm version bundled with node

mikeharder:
- In general, scripts should use the version of `npm` bundled with whatever version of `node` they are using

### DIFF
--- a/.sdkauto/initScript.sh
+++ b/.sdkauto/initScript.sh
@@ -2,8 +2,6 @@
 
 set -e
 
-npm install -g npm
-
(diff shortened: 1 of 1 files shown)


benbp (approved):

mikeharder:
@anthony-c-martin: This should fix the issue blocking check `SDK azure-resource-manager-schemas`.

anthony-c-martin (approved):

anthony-c-martin:
Thanks for fixing this!

PR #3022, opened 2024-03-29
Fix bug with generating schemas in windows with path separator.

Xynoclafe:
Path.sep had a bug in windows wherein it returned "\" as the separating character but the path itself had "/" as the character. As a result, generate commands didn't generate the schema files in windows. This PR fixes that.


anthony-c-martin (approved):
Thanks for fixing this!

PR #3056, opened 2024-04-19
Onboard all resource types to autorest v2

anthony-c-martin:
This PR deprecates the use of [autorest.azureresourceschema](https://github.com/Azure/autorest.azureresourceschema), and moves all schema generation onto [autorest.bicep](https://github.com/Azure/bicep-types-az/blob/main/src/autorest.bicep/).

Notable changes:
* There are many changes to the ordering of properties, but I've been using [`jdiff`](https://github.com/anthony-c-martin/dotfiles/blob/e092bdc0a6a49c0127869088784115efec3c86f2/.gitconfig#L26) to perform an ordering-insensitive diff. I haven't found anything which looks problematic.
* I've had to re-do the way post-processing works w.r.t additional schemas. This actually fixed Microsoft.Compute's generation, which has been silently failing for some time.
* I've had to remove a large number of tests that were highly structure-dependent. My opinion is that these are no longer adding value, since they are only testing old schema versions that were initially manually-authored.
* There are a number of schemas that appeared to be failing that are now being correctly generated.

Benefits:
* Schema generation is now down from ~5h30 to ~25mins!
* We are now running on a non-deprecated version of autorest.


generator/autorest.ts
```diff
@@ -2,15 +2,97 @@
...
+  // input-file:
+  // - path/to/file.json
+  // - path/to/other_file.json
+  // ```
+  for (const node of markdown.iterate(readmeMarkdown.markDown)) {
+    // We're only interested in yaml code blocks
+    if (node.type !== 'code_block' || !node.info || !node.literal ||
+      !node.info.trim().startsWith('yaml')) {
+      continue;
+    }
```
jorgecotillo, line 37: This section guarantees that we only read, right?: 

``` yaml $(tag) == 'package-bicep-2023-11'
input-file:
  - Microsoft.Resources/stable/2023-11-01/bicepClient.json
```

generator/autorest.ts
```diff
@@ -2,15 +2,97 @@
...
+
+  const filesByTag: Record<string, string[]> = {};
+  for (const file of inputFiles) {
+    const normalizedFile = normalizeJsonPath(file);
+    const match = pathRegex.exec(normalizedFile);
+    if (match) {
+      // Generate a unique tag. We can't process all of the different API versions in one autorest pass
+      // because there are constraints on naming uniqueness (e.g. naming of definitions), so we want to pass over
+      // each API version separately.
+      const tagName = `${match[2].toLowerCase()}-${match[3].toLowerCase()}`;
```
jorgecotillo, line 62: Not fully following this, I understand that we want to generate something unique but the why is not clear, is it because autorest chokes? because our code chokes? because the generated schema will have duplicates?

Also, what is the value of `match[2]` and `match[3]` how do you ensure these are unique?

generator/autorest.ts
```diff
@@ -2,15 +2,97 @@
...
+    } else {
+      console.warn(`WARNING: Unable to parse swagger path "${file}"`);
+    }
+  }
+
+  let generatedContent = `##Bicep
+
+### Bicep multi-api
+\`\`\`yaml $(bicep) && $(multiapi)
+${yaml.dump({ 'batch': Object.keys(filesByTag).map(tag => ({ 'tag': tag })) }, { lineWidth: 1000 })}
```
jorgecotillo, line 77: >{ lineWidth: 1000 } [](http://example.com/codeflow?start=79&length=19)

Curious about this number, is this a deliberate number to allow users to read the output? can this be anything? asking in case we need to change it later - for some reason - so we know where to make the update(s)

jorgecotillo:
>     }

Does autorest generates a tmpFolder? Why this check is after the cmd exec?

---
Refers to: generator/autorest.ts:101 in 293d23b. [](commit_id = 293d23b910e60b19847f32a7ea50e6555a79317f, deletion_comment = False)

jorgecotillo:
>     return await findRecursive(tmpFolder, p => path.extname(p) === '.json');

Is this code verifying that the file exists?

---
Refers to: generator/autorest.ts:103 in 293d23b. [](commit_id = 293d23b910e60b19847f32a7ea50e6555a79317f, deletion_comment = False)

generator/autorest.ts
```diff
@@ -2,15 +2,97 @@
...
+        filesByTag[tagName] = [];
+      }
+
+      filesByTag[tagName].push(normalizedFile);
+    } else {
+      console.warn(`WARNING: Unable to parse swagger path "${file}"`);
+    }
+  }
+
+  let generatedContent = `##Bicep
```
jorgecotillo, line 73: >##Bicep [](http://example.com/codeflow?start=26&length=7)

Is this just an arbitrary name? Or is a bicep readme what is being output?

generator/processors/helpers.ts
```diff
@@ -2,9 +2,79 @@
...
+  }
+
+  const parentDefinitions = findCyclicReferencingDefinitions(refVertices, definitionName);
+  for (const parentDefinition of parentDefinitions) {
+    replaceRefByName(schema.definitions[parentDefinition], definitionName, replacement);
+  }
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function doRecursively(schema: any, func: (input: any) => any) {
```
jorgecotillo, line 38: >doRecursively [](http://example.com/codeflow?start=9&length=13)

nit - `findDefinitionsRecursively` as the function name?

generator/autorest.ts
anthony-c-martin: Yes, exactly.

generator/autorest.ts
anthony-c-martin: Let me add a comment to explain better. Autorest groups files by "tag", and we're generating the config for autorest here. So the tags just need to be unique for grouping.

Here's an example of what the output format looks like:

    ##Bicep

    ### Bicep multi-api
    ```yaml $(bicep) && $(multiapi)
    batch:
      - tag: microsoft.securityinsights-2024-03-01
      - tag: microsoft.securityinsights-2024-01-01-preview
      ...

    ```
    ### Tag: microsoft.securityinsights-2024-03-01 and bicep
    ```yaml $(tag) == 'microsoft.securityinsights-2024-03-01' && $(bicep)
    input-file:
      - Microsoft.SecurityInsights/stable/2024-03-01/AlertRules.json
      - Microsoft.SecurityInsights/stable/2024-03-01/AutomationRules.json
      ...

    ```
    ### Tag: microsoft.securityinsights-2024-01-01-preview and bicep
    ```yaml $(tag) == 'microsoft.securityinsights-2024-01-01-preview' && $(bicep)
    input-file:
      - Microsoft.SecurityInsights/preview/2024-01-01-preview/AlertRules.json
      - Microsoft.SecurityInsights/preview/2024-01-01-preview/AutomationRules.json
      - Microsoft.SecurityInsights/preview/2024-01-01-preview/BillingStatistics.json
      ...

generator/autorest.ts
anthony-c-martin: I didn't change this - I'm guessing it's just a very large number to avoid cutting off lines when reading yaml. Perhaps autorest isn't able to handle wrapping of lines properly?

generator/autorest.ts
anthony-c-martin: Yes it's just an arbitrary name.

generator/autorest.ts
```diff
@@ -2,15 +2,97 @@
...
+      // because there are constraints on naming uniqueness (e.g. naming of definitions), so we want to pass over
+      // each API version separately.
+      const tagName = `${match[2].toLowerCase()}-${match[3].toLowerCase()}`;
+      if (!filesByTag[tagName]) {
+        filesByTag[tagName] = [];
+      }
+
+      filesByTag[tagName].push(normalizedFile);
+    } else {
+      console.warn(`WARNING: Unable to parse swagger path "${file}"`);
```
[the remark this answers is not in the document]
anthony-c-martin: Yes exactly. This means that we failed to parse the "input file" format from the readme (autorest configuration file).

For example, if the readme contained something like this:
```yml
input-file
- Fadsfosun # not a path to a .json file
```

generator/processors/helpers.ts
anthony-c-martin: Updated to `transformSchemaRecursively`.

anthony-c-martin:
> > ```
> > }
> > ```
> 
> Does autorest generates a tmpFolder? Why this check is after the cmd exec?
> 
> Refers to: generator/autorest.ts:101 in [293d23b](https://github.com/Azure/azure-resource-manager-schemas/commit/293d23b910e60b19847f32a7ea50e6555a79317f). [](commit_id = [293d23b](https://github.com/Azure/azure-resource-manager-schemas/commit/293d23b910e60b19847f32a7ea50e6555a79317f), deletion_comment = False)

The cmd exec will throw if there are any errors. I think this logic is just there to handle the case where the tmp directory was never created. It's not totally clear whether this is necessary - my guess is it'll happen if autorest doesn't generate any files, to avoid `findRecursive` from throwing.

jorgecotillo (approved):

PR #3182, opened 2024-08-08
Add new parameter to saveAutoGeneratedSchemaRefs

Xynoclafe:
Add new parameter to saveAutoGeneratedSchemaRefs to remove stale entries from autogeneratedResources


anthony-c-martin (approved):
Thanks for fixing this!

PR #3301, opened 2025-01-13
Remove test server

Xynoclafe:
This PR removes the server.ts file and the associated test server functionality.


anthony-c-martin:
You'll also need to update https://github.com/Azure/azure-resource-manager-schemas/blob/main/tools/package-lock.json. You can do this by cd-ing to the tools directory and running `npm i`:

```sh
cd tools
npm i
```

anthony-c-martin (approved):

PR #3314, opened 2025-01-28
Fix ICU build issues

Xynoclafe:
Fixes build issues caused by a breaking change in the base ubuntu image… wrt ICU packages - https://github.com/actions/runner-images/issues/10989
Also fixes a typo in awsconnector type generation and fixes a potential issue in node spawn command.


anthony-c-martin (approved):
Thanks for fixing this!

PR #3353, opened 2025-03-04
Add updated deployment parameters schema 2025-03-01

kalbert312:
Adds a new version of the deployments template parameters file schema that includes extensionConfigs support.


schemas/2025-03-01/tenantDeploymentParameters.json
```diff
@@ -0,0 +1,102 @@
...
+      "additionalProperties": {
+        "$ref": "#/definitions/extensionConfigItem"
+      },
+      "description": "A template extension's configuration. The keys of this object are defined by the extension's configuration schema."
+    },
+    "extensionConfigItem": {
+      "type": "object",
+      "properties": {
+        "value": {
+          "$ref": "https://schema.management.azure.com/schemas/2019-08-01/tenantDeploymentTemplate.json#/definitions/parameterValueTypes",
```
anthony-c-martin, line 79: Should this be a relative path?

schemas/2025-03-01/deploymentParameters.json
```diff
@@ -0,0 +1,102 @@
+﻿{
+  "id": "https://schema.management.azure.com/schemas/2025-03-01/deploymentParameters.json#",
```
anthony-c-martin, line 2: I know this is probably not the intended usage, but we typically add to the existing schemas rather than add new ones. (assuming we're not making breaking changes, which we generally would not do anyway).

The problem is that we have no mechanism to actually point people to use new schemas when they are published.

PR #3354, opened 2025-03-05
Add extensionConfigs property to deployment parameters schema.

kalbert312:
Add new optional `extensionConfigs` property to the deployment parameters file schema. No breaking changes.

### DIFF
--- a/schemas/2019-04-01/deploymentParameters.json
+++ b/schemas/2019-04-01/deploymentParameters.json
@@ -19,6 +19,13 @@
         "$ref": "#/definitions/parameter"
       },
       "description": "Collection of parameters to pass into a template"
+    },
+    "extensionConfigs": {
+      "type": "object",
+      "additionalProperties": {
+        "$ref": "#/definitions/extensionConfig"
+      },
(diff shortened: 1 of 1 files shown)


schemas/2019-04-01/deploymentParameters.json
```diff
@@ -53,6 +60,38 @@
...
+    },
+    "extensionConfigItem": {
+      "type": "object",
+      "properties": {
+        "value": {
+          "$ref": "#/definitions/extensionConfigItemValueTypes",
+          "description": "The extension config property value. The type of the value is defined by the extension's configuration schema."
+        },
+        "keyVaultReference": {
+          "$ref": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#/definitions/keyVaultReference",
```
kalbert312, line 79: Usage of URL here is based on existing usages for key vault reference.

kalbert312:
There's a tenant, sub, and mgmt group variant of this schema. They are not used by Bicep as far as I can tell. Should those be updated too or should we move towards only 1 schema?

anthony-c-martin (approved):

PR #3624, opened 2026-01-06
Clear autorest temp directories after every basepath

Xynoclafe:
Cleans up autorest temp directory during schema generation. This should make encountering disk space issues less likely


anthony-c-martin (approved):
Great find!

PR #3632, opened 2026-01-13
Generate schemas straight from Bicep types

anthony-c-martin:
This PR was mostly generated using AI. Steps:
* Copy repo structure from https://github.com/Azure/template-reference-generator (which is also generating artifacts based on bicep types)
* Ask Copilot to create test framework with baselines
* Multiple iterations of re-running generation + tests until generation reaches the expected output format

Since it's a complete replacement for the existing code in this repo, it's probably easiest to review by just looking through the code that is net new:
* [src](https://github.com/Azure/azure-resource-manager-schemas/tree/ant/from_types/src) - the C# generation source code and tests
* [.github](https://github.com/Azure/azure-resource-manager-schemas/tree/ant/from_types/.github) - GH Actions


.github/workflows/generate.yml
```diff
@@ -0,0 +1,57 @@
+name: Generate Schemas
+
```
Alberto3903: .editorconfig

Alberto3903 (changes requested):

anthony-c-martin:
Split into separate PRs

PR #3677, opened 2026-03-20
Add optional ability to generate schemas from Bicep types

anthony-c-martin:
This PR introduces an alternative ARM JSON schema generation path using a new C# tool (`TemplateSchemaGenerator`) that reads from the bicep-types-az generated types, instead of the existing TypeScript/AutoRest/Swagger pipeline.

### What's new

**New C# project: TemplateSchemaGenerator**

A self-contained CLI tool that:
- Reads resource types from the generated folder
- Converts Bicep type definitions to JSON Schema format, including correct scope-based resource definition keys (`resourceDefinitions`, `subscription_resourceDefinitions`, etc.)
- Handles circular reference detection with a targeted list of known cycle-break paths per provider namespace
- Writes per-provider schema files (`{apiVersion}/{ProviderNamespace}.json`) to an output folder
- Optionally regenerates the `common/autogeneratedResources.json` root schema index
- Supports full-generation (with output cleanup) and single-provider incremental runs

Key components:
- `JsonSchemaGenerator.cs` — core Bicep-to-JSON-Schema conversion
- `MainGenerator.cs` — orchestrator: groups types by provider/API version, manages output folder lifecycle
- `ResourceTypeProvider.cs` — loads resource types from the bicep-types-az generated folder
- `AutogeneratedResourcesGenerator.cs` — regenerates the root schema reference index
- `Program.cs` — CLI entry point (`--source-folder`, `--output-folder`, `--provider-namespace`, `--no-update-root-schemas`, `--verbose`)

**TypeScript integration**

- `autogenlist.ts`: Added `csharpGeneratorEnabledProviders` (currently empty). Providers added here are excluded from the AutoRest-based pipeline.
- `generate.ts`: Added `generateSchemasWithCsharpGenerator()` — invokes the compiled C# binary, discovers its output, and translates results into `SchemaConfiguration` entries compatible with the existing pipeline.
- `generateall.ts`: Runs the C# generator first (before the AutoRest loop), for any provider in `csharpGeneratorEnabledProviders`.
- `generatesingle.ts`: Added `--provider-namespace` option to invoke single-provider C# generation from the CLI.

### Migration approach

The two pipelines coexist. To migrate a provider to the C# generator, add its namespace to `csharpGeneratorEnabledProviders` in `autogenlist.ts`. Until then, all existing behavior is unchanged.

### Tests

Comprehensive test coverage in `TemplateSchemaGenerator.Tests`:
- `JsonSchemaGeneratorTests` — unit tests for schema generation logic with baseline JSON fixtures for several providers (Batch, Compute, ContainerService, CosmosDB, DataFactory, DataShare, EventGrid, Resources, Web)
- `ExportTemplateEngineScenarioTests` — end-to-end export scenario tests
- `OutputPreservationTests` — verifies that hand-authored files (deployment templates, CreateUIDefinition, policy definitions, etc.) are preserved during full-generation cleanup


src/TemplateSchemaGenerator/Generators/JsonSchemaGenerator.cs
```diff
@@ -0,0 +1,826 @@
...
+
+                oneOf.Add(new JsonObject
+                {
+                    ["type"] = "object",
+                    ["properties"] = properties,
+                    ["required"] = required,
+                });
+            }
+
+            var resourceRootProperties = new JsonObject
```
Xynoclafe: We may be stripping away and not adding back properties like "name" here in some case

Xynoclafe (approved):

PR #3692, opened 2026-04-08
Remove TLS1_3 from Microsoft.Storage minimumTlsVersion enum

sethvargo:
Discussed with the PMs. This is not a breaking change because the functionality does not exist (currently returns an error).

/cc @brendandburns

### DIFF
--- a/schemas/2023-05-01/Microsoft.Storage.json
+++ b/schemas/2023-05-01/Microsoft.Storage.json
@@ -3878,8 +3878,7 @@
               "enum": [
                 "TLS1_0",
                 "TLS1_1",
-                "TLS1_2",
-                "TLS1_3"
+                "TLS1_2"
               ],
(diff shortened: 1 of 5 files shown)


sethvargo:
@microsoft-github-policy-service agree

I'm a Microsoft employee :)

anthony-c-martin:
@sethvargo - the files in this repo are automatically generated from Resource Provider OpenAPI specs (https://github.com/Azure/azure-rest-api-specs). If you manually update them, they'll be reverted the next time generation runs.

The correct place to make this fix is in the source repo - e.g. here: https://github.com/Azure/azure-rest-api-specs/blob/4accdfe81c521544ed37d3b53a1f3de3391cdf9d/specification/storage/resource-manager/Microsoft.Storage/stable/2023-05-01/storage.json#L3251-L3264

PR #3709, opened 2026-04-10
Enhance C# schema generation capabilities

anthony-c-martin:
Introduce a new option to generate only C# schemas and update the list of enabled providers. Adjustments made to schema path handling and dependencies ensure improved functionality and maintainability.


Xynoclafe (approved):

.github/workflows/main.yml
```diff
@@ -44,11 +41,6 @@ jobs:
       - name: Run CI tests
         run: npm test
         working-directory: ./tools
-
-      - name: Deployments Schema tests
```
Xynoclafe, line 48: I assume this change doesn't remove these tests, but rather that the test doesn't work this way anymore due to the new changes?

anthony-c-martin: The test project is moved under the main .sln file, so the tests are running under the ".NET Tests" CI job instead

Xynoclafe (approved):

PR #3713, opened 2026-04-15
Onboard Microsoft.DevHub for C# generation

anthony-c-martin:


schemas/2024-08-01-preview/Microsoft.DevHub.json
```diff
@@ -380,7 +378,6 @@
       "type": "object"
     },
     "Build": {
-      "description": "Build details of the repository associated with the workflow.",
```
Xynoclafe, line 383: Are all these description deletes valid

anthony-c-martin: Yes - descriptions aren't used in export

schemas/2024-05-01-preview/Microsoft.DevHub.json
```diff
@@ -65,6 +65,7 @@
         }
       },
       "required": [
+        "location",
```
Xynoclafe, line 68: Why do we have a required location property now that we didn't have before? I don't see any changes in the underlying spec

anthony-c-martin: Required location is correct - the previous behavior was incorrectly counting it as optional. The resource definition extends `TrackedResource`, which has location marked as required: https://github.com/Azure/azure-rest-api-specs/blob/5142fed9fc2b10464d3d45eb64edd252731974db/specification/developerhub/resource-manager/Microsoft.DevHub/DeveloperHub/preview/2024-05-01-preview/iac.json#L450

Xynoclafe (approved):

PR #3716, opened 2026-04-16
Add config mechanism to avoid merge conflicts

anthony-c-martin:
This has no functional impact on behavior since the default config values are all disabled. By having a pre-populated list rather than an empty one, this means multiple onboarding PRs can be raised simultaneously without each one running into merge conflicts.


generator/autogenlist.ts
```diff
@@ -21,8 +21,180 @@ import { postProcessor as awsConnectorPostProcessor } from './processors/Microso
...
 
-// Providers that have been onboarded to the C# generator. Any provider in this list must be be excluded from autogenList.
-export const csharpGeneratorEnabledProviders: string[] = [];
+type CsharpGeneratorConfig = {
+  namespace: string,
+  enabled: boolean,
+}
+
+// Providers that have been onboarded to the C# generator instead of autorest.
+const csharpGeneratorProviders: CsharpGeneratorConfig[] = [
```
Xynoclafe, line 30: Is this a list of ALL RPs? Or just the few with configurations for them later in this file?
The reason I ask is that there are some RPs not mentioned in this file that are generated. I am assuming that ANY RP being added to the new C# generation would be added to this list here?

anthony-c-martin: Yeah exactly, just RPs with configs later in this file - I didn't spend a long time gathering the names. It doesn't need to be a complete list because functionally it's the same to be in the list and disabled, and omitted from the list.

Xynoclafe (approved):

PR #3720, opened 2026-04-16
Onboard Microsoft.Maps for C# generation

anthony-c-martin:
Enables `Microsoft.Maps` in the `csharpGeneratorProviders` list so schemas are generated via the C# generator instead of AutoRest.

12 schema files regenerated across all existing API versions.

### DIFF
--- a/schemas/2017-01-01-preview/Microsoft.Maps.json
+++ b/schemas/2017-01-01-preview/Microsoft.Maps.json
@@ -18,7 +18,7 @@
           "type": "string"
         },
         "name": {
-          "description": "The name of the Maps Account.",
+          "description": "The resource name",
           "type": "string"
         },
         "sku": {
@@ -55,7 +55,9 @@
         }
       },
       "required": [
+        "location",
         "name",
+        "sku",
         "apiVersion",
         "type"
       ],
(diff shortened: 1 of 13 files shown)


schemas/2021-02-01/Microsoft.Maps.json
```diff
@@ -181,13 +183,11 @@
       "type": "object"
     },
     "MapsAccountProperties": {
-      "description": "Additional Map account properties",
       "properties": {
         "disableLocalAuth": {
           "description": "Allows toggle functionality on Azure Policy to disable Azure Maps local authentication support. This will disable Shared Keys authentication from any usage.",
           "oneOf": [
             {
```
Xynoclafe, line 190: Is removal of a default expected?

schemas/2017-01-01-preview/Microsoft.Maps.json
```diff
@@ -55,16 +55,24 @@
         }
       },
       "required": [
+        "location",
```
Xynoclafe, line 58: Where are these new required fields coming from?

anthony-c-martin: Required location is correct - the previous behavior was incorrectly counting it as optional. The resource definition has location marked as required: https://github.com/Azure/azure-rest-api-specs/blob/5142fed9fc2b10464d3d45eb64edd252731974db/specification/maps/resource-manager/Microsoft.Maps/Maps/stable/2017-01-01-preview/maps-management.json#L626

schemas/2021-02-01/Microsoft.Maps.json
anthony-c-martin: Yes - defaults aren't represented in Bicep types, and they aren't used by Export

Xynoclafe (approved):

PR #3730, opened 2026-04-22
Onboard some other providers for C# generation

anthony-c-martin:


codeinputmachine:
Hey @anthony-c-martin, Code Input detected this PR has a merge conflict. The conflicts of this PR can be resolved with a semantic merge driver. Code Input can do that automatically: https://codeinput.com/r/3eJXh7Ht6jq Let me know if you need more help with this conflict or how Code Input works.

schemas/2016-02-03/Microsoft.Devices.IoTHub.json
Xynoclafe, line 1: I am assuming that the IoThub schemas go folded into the larger Devices schema?
So what happens here if lets say 2 different resource types share an API version within a single RP? The autorest generation handled them by adding these sufffixes. What happens here? Is it generated in such a way that the file contains both schemas and one is not overwriting the other?

anthony-c-martin: Yes exactly - everything is folded together, and the suffixes are no longer necessary.

The Bicep types index (which this information is sourced from) groups everything by provider namespace, so the problem goes away entirel.

Xynoclafe (approved):

Xynoclafe (approved):

gary-x-li (approved):

PR #3821, opened 2026-07-23
Onboard more providers to C# generation

anthony-c-martin:


generator/package.json
```diff
@@ -38,5 +38,8 @@
     "ts-node": "^10.9.2",
     "typescript": "^6.0.3",
     "yargs": "^18.0.0"
+  },
+  "allowScripts": {
+    "autorest@3.8.0": true
   }
```
Xynoclafe, line 44: What does this do?

anthony-c-martin: See https://github.blog/changelog/2026-06-09-upcoming-breaking-changes-for-npm-v12/

Older versions of NPM would automatically run scripts as part of a package install, which is a massive potential security hole when you think about it, if someone is able to slip in a malicious script.

In newer versions of NPM, this functionality defaults to being disabled, but with an allowlist (this setting). Autorest depends on this functionality, and is a Microsoft-published package, so I've opted it in.

Xynoclafe (approved):

PR #3836, opened 2026-08-27
Disable template schema autogen for Microsoft.ProviderHub

msmbaldwin:
## Summary

Sets `disabledForAutogen: true` for `Microsoft.ProviderHub` in `generator/autogenlist.ts`.

## Why

Microsoft.ProviderHub is an internal resource provider. Per this repo's stated policy:

> We only publish template schemas for resource providers that are **publicly available**. This means that there should be no restrictions (private preview, internal-only allowlisting) on who can call your APIs.

ProviderHub does not meet that bar, so its ARM/Bicep template reference schemas should not be published.

## Effect

- Stops generation of the `learn.microsoft.com/azure/templates/microsoft.providerhub/**` reference pages.
- Removes Microsoft.ProviderHub from ARM/Bicep template IntelliSense.

## Change

Single flag added to the existing ProviderHub entry, consistent with the ~13 providers already carrying `disabledForAutogen: true` (e.g. Microsoft.Logic, Microsoft.AgFoodPlatform).

### DIFF
--- a/generator/autogenlist.ts
+++ b/generator/autogenlist.ts
@@ -1058,6 +1058,7 @@ const autoGenList: AutoGenConfig[] = [
     {
         basePath: 'providerhub/resource-manager',
         namespace: 'Microsoft.ProviderHub',
+        disabledForAutogen: true,
         postProcessor: providerHubPostProcessor
     },
     {


msmbaldwin:
@Xynoclafe @anthony-c-martin - could one of you review when you have a moment? This sets `disabledForAutogen` on Microsoft.ProviderHub, an internal resource provider, in line with this repo's policy of only publishing template schemas for publicly available providers. All checks are green. Thanks!

anthony-c-martin:
@msmbaldwin the schemas in this repo are generated from Azure REST API specs that have been published by service teams (https://github.com/Azure/azure-rest-api-specs)

The correct way to handle this is to contact the owners of https://github.com/Azure/azure-rest-api-specs/tree/main/specification/providerhub/resource-manager/Microsoft.ProviderHub/ProviderHub, and ask them to redact it. Once that's been done, this'll be cleaned up automatically.

EDIT: Looks like this is already being taken care of: https://github.com/Azure/azure-rest-api-specs/pull/45913

PR #3849, opened 2026-09-05
Onboard more providers to C# generation

anthony-c-martin:
* Onboard many more RPs for C# schema generation
* Invert the list of configured RPs from an allowlist to a blocklist, so that RPs are onboarded by default


generator/autogenlist.ts
```diff
@@ -20,211 +20,21 @@ import { postProcessor as awsConnectorPostProcessor } from './processors/Microso
...
-    { namespace: 'Microsoft.Web', enabled: false },
-    { namespace: 'Microsoft.WindowsESU', enabled: true },
-    { namespace: 'Microsoft.WindowsIoT', enabled: true },
-    { namespace: 'Microsoft.WorkloadMonitor', enabled: true },
-    { namespace: 'Microsoft.Workloads', enabled: true },
-    { namespace: 'Nginx.NginxPlus', enabled: true },
-    { namespace: 'Oracle.Database', enabled: true },
-    { namespace: 'Providers.Test', enabled: true },
+// Providers that have not been onboarded to the C# generator and still use AutoRest.
+export const csharpGeneratorDisabledProviders: string[] = [
```
Xynoclafe, line 24: Are we okay onboarding Resources to this?

anthony-c-martin: Assuming you mean `Microsoft.Resources` - they're actually already onboarded: https://github.com/Azure/azure-resource-manager-schemas/blob/901a4af091372f59b067f20621c0ecdfe44b187c/generator/autogenlist.ts#L183

Xynoclafe: Huh - I would've assumed that would have had the potential to cause problems. Makes sense

Xynoclafe (approved):

